PR #25, opened 2017-11-08
Pull font weight, tokenise + global

siddharthkp:


landitus:
Great! I think `medium` would suit better than `semibold`, as it's much easier to remember and sits between normal and bold as well. For us it's just an abstraction of the weight related to the others.

I'm looking at this font-weight reference for how type foundries classify the weights. We should change the values both in the font face import and in the token, like this:
 
100    Extra Light or Ultra Light
200    Light or Thin
300    Book or Demi
**400    Normal or Regular**
**500    Medium**
600    Semibold, Demibold
**700    Bold**
800    Black, Extra Bold or Heavy
900    Extra Black, Fat, Poster or Ultra Black

Another reference is Adobe Typekit:
screen shot 2017-11-09 at 12 11 47

siddharthkp:
ooh, good catch!

PR #31, opened 2017-11-22
[WIP] New docs

siddharthkp:
New docs 

- [x] Use a router to keep all the folders on the same instance
- [x] Use metadata from the components 
- [ ] Use markdown for documentation


siddharthkp:
Merging this even though it's WIP coz it's more expensive to keep this around 😅

Will make smaller pull requests from now on!

Issue #40, opened 2017-12-11
How customisable should components be?


siddharthkp:
It's easy to add customisations to components but each prop adds complexity, they will be really hard to remove later on.

Opinion: We should pick consistency over flexibility.

Example:  We have 2 kind of inputs that have similar behaviour: `<Input disabled />` and `<Input readonly />`,

_Rigid_ components will help make it easier to test and maintain components. This can feel limiting at first, but it will help us build consistent UIs and remove some of the decision fatigue while picking components. (Bonus: easier to test + maintain)

-- 

Another interesting way to look at this is:

- Build flexible atomic components
- Build molecules with strict props, which is essentially locking down props of underlying atoms to expose a simpler building block for others.

PR #155, opened 2018-01-05
Implement EmptyState component, add color property to Icon

### LINKED ISSUE #141: Empty State Component
Since empty space is confusing, this component can be used instead when a list is empty. It could also be used for other "non-ideal" state, like errors.

screen shot 2018-01-04 at 12 52 02 pm

nkohari:
Closes #141.

### DIFF
--- a/src/components/atoms/empty-state.js
+++ b/src/components/atoms/empty-state.js
@@ -1,0 +1,42 @@
+import React from 'react'
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+import Icon, { StyledIcon } from './icon'
+import { colors } from '../../tokens'
+
+const Wrapper = styled.div`
+  width: 100%;
+  height: 100%;
+  text-align: center;
+  ${StyledIcon} {
+    line-height: 1em;
+  }
+`
+
+const Content = styled.div`
+  margin-top: 1em;
+`
+
+const EmptyState = props => {
+  return (
+    <Wrapper>
+      <Icon type={props.icon} size={props.iconSize} color={props.iconColor} />
+      <Content>{props.children}</Content>
+    </Wrapper>
+  )
+}
+
+EmptyState.displayName = 'EmptyState'
+
+EmptyState.propTypes = {
+  icon: PropTypes.string.isRequired,
+  iconSize: PropTypes.number,
+  iconColor: PropTypes.string
+}
+
+EmptyState.defaultProps = {
+  iconSize: 100,
+  iconColor: colors.grayMedium
+}
+
+export default EmptyState
--- a/src/components/atoms/empty-state.js
+++ b/src/components/atoms/empty-state.js
@@ -2,7 +2,7 @@ import React from 'react'
 import styled from 'styled-components'
 import PropTypes from 'prop-types'
 import Icon, { StyledIcon } from './icon'
-import { colors } from '../../tokens'
+import { colors, spacing } from '../../tokens'
 
 const Wrapper = styled.div`
   width: 100%;
@@ -14,7 +14,7 @@ const Wrapper = styled.div`
 `
 
 const Content = styled.div`
-  margin-top: 1em;
+  margin-top: ${spacing.xsmall};
 `
 
 const EmptyState = props => {
--- a/src/components/atoms/empty-state.js
+++ b/src/components/atoms/empty-state.js
@@ -20,7 +20,7 @@ const Content = styled.div`
 const EmptyState = props => {
   return (
     <Wrapper>
-      <Icon type={props.icon} size={props.iconSize} color={props.iconColor} />
+      <Icon type={props.icon} size={100} color={colors.grayMedium} />
       <Content>{props.children}</Content>
     </Wrapper>
   )
@@ -29,14 +29,7 @@ const EmptyState = props => {
 EmptyState.displayName = 'EmptyState'
 
 EmptyState.propTypes = {
-  icon: PropTypes.string.isRequired,
-  iconSize: PropTypes.number,
-  iconColor: PropTypes.string
-}
-
-EmptyState.defaultProps = {
-  iconSize: 100,
-  iconColor: colors.grayMedium
+  icon: PropTypes.string.isRequired
 }
 
 export default EmptyState
--- a/src/components/atoms/empty-state.md
+++ b/src/components/atoms/empty-state.md
@@ -2,22 +2,8 @@ The `icon` property is required, and should be set to a valid icon type. (Same a
 property in the `<Icon>` component.) Any children of the component will be rendered underneath
 the icon.
 
-```js
-<EmptyState icon="clients">
-  <Text>No items have been defined.</Text>
-</EmptyState>
-```
-
-Icon size and color can be configured using the `iconSize` and `iconColor` properties, respectively.
-
-```js
-<EmptyState icon="clients" iconSize={60}>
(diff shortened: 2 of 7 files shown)


siddharthkp:
I have a few design questions about this, i noticed there are 3 props here.

`icon` makes sense to me, but should we make size and color customisable? This sounds like something that should be consistent across screens.

Related: https://github.com/auth0/cosmos/issues/40

```js
{
  icon: PropTypes.string.isRequired,
  iconSize: PropTypes.number,
  iconColor: PropTypes.string
}
```

src/components/atoms/empty-state.md
```diff
@@ -0,0 +1,23 @@
+The `icon` property is required, and should be set to a valid icon type. (Same as the `type`
+property in the `<Icon>` component.) Any children of the component will be rendered underneath
+the icon.
+
```
siddharthkp: check out the source of [Input documentation](https://github.com/auth0/cosmos/blob/master/src/components/atoms/input.md) to see how to add the props table, I'll add it to the example [`_box` component](https://github.com/auth0/cosmos/blob/master/src/components/atoms/_box.js) as well and add more to the [contributing guide](https://github.com/auth0/cosmos/blob/master/contributing.md).

nkohari: Ah, thanks! I think I just copied (I think) the `<Tabs>` component, and didn't realize there was a table like that available. I'll update the docs.

src/overview/src/empty-states.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+
+import Container from './container'
+import { EmptyState, Text } from '../../components'
+
+const EmptyStates = () => (
+  <Container title="Empty State">
+    <EmptyState icon="clients">
+      <Text>No items have been added to this section.</Text>
```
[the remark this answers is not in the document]
nkohari: Agreed, it felt strange as I was doing it. I'm not sure if this is a great idea, but we could add a class like `cosmos-component` to all classes, and apply the CSS to `<span>`s within. Not sure if that would feel more or less hacky than requiring `<Text>`.

src/components/atoms/empty-state.js
```diff
@@ -0,0 +1,42 @@
...
+const EmptyState = props => {
+  return (
+    <Wrapper>
+      <Icon type={props.icon} size={props.iconSize} color={props.iconColor} />
+      <Content>{props.children}</Content>
+    </Wrapper>
+  )
+}
+
+EmptyState.displayName = 'EmptyState'
```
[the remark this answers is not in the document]
nkohari: It seemed like it was using the filename to infer the display name, and assigning it `Empty-state`, so this was my attempt to force it to use `EmptyState` instead. :)

src/components/atoms/empty-state.md
nkohari: I noticed a few issues using this:

screen shot 2018-01-05 at 4 13 41 pm

1. Using quotes around the default value will pull those quotes into the prop value. (Using double quotes actually causes a syntax error to be displayed in the docs.)
2. I'm not sure we really want a default value for `icon`, but I get an error message if I don't supply one.
3. Maybe the `colors.grayMedium` token should be resolved to its hex value? It's definitely confusing since it appears here as a string.

landitus:
I think the Empty State is more of an organism, and not an atom, as it has multiple elements that can work separately.

src/components/atoms/empty-state.md
siddharthkp: 1. Ouch, let me see what I can do, value handling in the prop table is pretty fragile right now.

2. You can get around that easily by changing your code in the example.

```jsx
<EmptyState icon="help">...</EmptyState>
``

3. I'm still not sure if this is the right way to go: https://github.com/auth0/cosmos/pull/155#issuecomment-355660219

**Comment by siddharthkp** [comment]
> I think the Empty State is more of an organism, and not an atom, as it has multiple elements that can work separately.

*molecule? 😋


@landitus open question for you: https://github.com/auth0/cosmos/pull/155#issuecomment-355660219

**Comment by siddharthkp** [comment]
Also, a tiny nit pick, can merge it without this as well: What should we call this component?

`EmptyState` is confusing because of `state` in React components. Is `Placeholder` or `NoContent` a better name?

**Comment by nkohari** [comment]
Blueprint calls this `NonIdealState`. Not sure if that's better, though. :)

**Comment by nkohari** [comment]
Just added a commit to remove the extra configuration. It also has a small patch that fixes the check in the prop table when there is no default value for a given property.

File: src/components/atoms/empty-state.js [hunk 4]
```diff
@@ -0,0 +1,35 @@
...
+  width: 100%;
+  height: 100%;
+  text-align: center;
+  ${StyledIcon} {
+    line-height: 1em;
+  }
+`
+
+const Content = styled.div`
+  margin-top: 1em;
```
siddharthkp: Whoops, sorry I didn't notice this earlier. `em` is relative, we're trying to lock the values in by using [_spacing tokens_](https://github.com/auth0/cosmos/blob/master/src/tokens/spacing.js)

nkohari: Whoops! I'll fix that.

src/components/atoms/empty-state.js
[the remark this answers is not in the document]
siddharthkp: Let me look into it, probably a bug on the tooling, you shouldn't have to do this 😄

src/overview/src/empty-states.js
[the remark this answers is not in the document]
siddharthkp: Do we need a class? Would it not take the global font-size? (unless overriden)

I'm not sure what to do with this right now, so i'll open an issue and merge this for now ✅

src/components/atoms/empty-state.js
nkohari: Added a new commit that fixes this. I left the `line-height: 1em` for the icon image, since the purpose there is really just to treat the icon as an image rather than an in-line glyph.

nkohari:
Hey, sorry, I thought I pushed this commit but I guess I didn't! :|

landitus:
- Ok, sounds good to be place under `molecules`. But because the design has more components involved, like the title and the actions I looks more complex. I don't yet know what's the limit to be organism (and if we need to get into that issue).
- What about adding the main title and the actions (buttons), like in the design mockup? Check https://github.com/auth0/cosmos/issues/141 Will these be props or children code? Or this component will have a parent component? I do see this component as a replacement of the whole section content. As such, it has some layout involved as well, the title and the actions. wdyt?

screen shot 2018-01-09 at 12 00 29

nkohari:
It should be possible to add those other components as children of the `<EmptyState>`, like:

```js
<EmptyState>
  <Text>No items have been added to this section.</Text>
  <Link ... />
  <Button ... />
</EmptyState>
```

I figured this was more flexible than adding several optional props to the `<EmptyState>` itself. If we need to tweak the style so the other items are laid out in a different way, we can do that. That seem ok?

landitus:
From the design perspective, this component is not very flexible, and for the most part has very strict layout. I think the text (with link) part can be, but all the other elements are either required and or positioned in a specific way. In these situations I'm more inclined to have the layout and elements all taken care by the component (and avoid flexibility)

nkohari:
Sure, if the component will display those same elements every time, it makes sense to build them all in. I'll make some changes to it.

landitus:
Sounds good!

PR #167, opened 2018-01-10
Composite empty state

nkohari:
This implements the feedback from #155. This patch also includes adding support for `Button` components with both icons and text, which was necessary to implement `EmptyState`.

screen shot 2018-01-10 at 2 50 01 pm


siddharthkp:
We have a deployed instance for this PR now 🎉 

https://auth0-cosmos-composite-empty-state.now.sh

PR #170, opened 2018-01-11
Add now-cd

siddharthkp:
- [ ] Need new token!


siddharthkp:
There's something wrong with the token 😢 

> You are on the OSS plan.

PR #171, opened 2018-01-11
Packaging part 1

siddharthkp:
- [x] Each component is inside a directory
- [x] Change paths to accommodate for this change
- [x] changed doc tooling to handle this change
- [x] hide internal components (that start with `_`)
- [x] changed `ActionInput` into an internal component (`_ActionInput`)

A component directory looks like this:

```
- atoms
  - button
    - index.js
    - button.md
  - input
    - index.js
    - input.md
```

Naming it `index.js` keeps imports the same.
`import button from 'button'` and not `import button from 'button/button'`

Ran `npm run sceenshot` and didn't get any visual diff


landitus:
Nice! 

I think I've missed why `ActionInput` is a hidden component. Aren't we exposing it to the docs?

nkohari:
Since this moves several things around, I'm just going to merge it so we can avoid more difficult merges later. If we want to make other changes we can do that in another PR.

landitus:
Ok

PR #172, opened 2018-01-11
Dialed back on the design changes to reflect the actual design

landitus:
I made some changes to colors, heights and elements to look as the do today in production code. This will help if we use it to test in an actual app, by blending in. It also gives us the opportunity to demo visual changes with our system tokens, as the new way to move forward with design changes. Not everything is 100% accurate to production, as some elements have been improved. 

Still, buttons colors are broken as I didn't use tokens properly. I meant to wait for the merge of the tokens PR.


siddharthkp:
Okay there are too many design decisions here that are becoming difficult to merge, I'd suggest copying the changes from a new branch from master

Issue #117, opened 2018-01-04
Tooltip Component


nkohari:
![screen shot 2018-01-04 at 11 33 10 am](https://user-images.githubusercontent.com/1576/34574239-d0601b6c-f144-11e7-8a30-212b185ed943.png)

PR #178, opened 2018-01-12
Add tooltip

beneliflo:
Implementation for https://github.com/auth0/cosmos/issues/117


src/components/atoms/tooltip/tooltip.md
```diff
@@ -0,0 +1,19 @@
+```js props
+<Tooltip {props} />
```
siddharthkp: Can you also add some text to the tooltip, the example looks incomplete without it 😄 

This should work:
```jsx
<Tooltip {props}>Here is some text</Tooltip>
```

src/components/atoms/tooltip/tooltip.md
```diff
@@ -0,0 +1,19 @@
...
+```
+
+---
+
+#### Examples
+
+Tooltip top:
+
+```js
+<Tooltip top />
```
siddharthkp: Should we also add the thing tooltip is on? I'm not sure how will the components align in that case.

landitus:
Maybe we could do something like:
<Tooltip top><Icon ... /></Tooltip>

src/components/atoms/tooltip/tooltip.md
landitus: Yes, maybe we could have:
`<Tooltip top><Icon ... /></Tooltip>

siddharthkp: Higher order component! I like that

src/components/atoms/tooltip/index.js
```diff
@@ -0,0 +1,56 @@
...
+const StyledTooltip = styled.div`
+  background: ${colors.tooltip.background};
+  color: ${colors.tooltip.text};
+  border-radius: ${misc.radius};
+  max-width: 200px;
+  text-align: center;
+  position: relative;
+  padding: ${spacing.xsmall};
+  line-height: normal;
+  font-size: 13px;
```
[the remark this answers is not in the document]
beneliflo: -`position: relative` it isn't of constant use. It is normally used when you have an element inside with `position: absolute`. I think it is not necessary to add it as a component in this case.
-`line-height: normal` I had to use this for text alignment mostly. This we can add it as a token :)

What you said with this should come from `tokens/fonts`? Build an array for specific component like `tokens/colors` or separate css properties like now on `tokens/fonts`?

siddharthkp:
Looks pretty good!!

Tiny nitpicks:

Should the API be `top` and `bottom` or `above` and `below`, what's more intuitive?

The example in `overview` might be broken:

screen shot 2018-01-17 at 8 02 59 pm

landitus:
I think for positioning we should use `top` and `bottom`. 
BTW, we should add a good width limit to accomplish a good reading experience. 200px could work for now. Check this bootstrap example:

screen shot 2018-01-17 at 11 49 10

beneliflo:
@siddharthkp yep I forgot update the last changes in `overview`. Fixing...
@landitus It is working with 200px at the moment :). Missing add the right side and the left side too.

siddharthkp:
Looks perfect. Merged!

Issue #128, opened 2018-01-04
Icon Component


nkohari:
![screen shot 2018-01-04 at 12 27 53 pm](https://user-images.githubusercontent.com/1576/34576040-b30e3fe8-f14a-11e7-8576-aa276b274d2f.png)

siddharthkp:
Some context here: https://github.com/auth0/cosmos/issues/58#issuecomment-351789268

PR #181, opened 2018-01-15
First implementation of SVG icons

nkohari:
This patch adds support for a new [Icon component](https://github.com/auth0/cosmos/issues/128) that returns SVGs instead of using the icon font.

As you might know, working with SVGs in React can be a little strange, especially if you're trying (as we are) to bundle the SVGs as part of a library. For our implementation, I chose to parse out the path data from SVG files into a lookup table, and then write out an `<svg>` element using JSX that contains the correct path(s) from the lookup table.

To do this, there is a new build script called `tooling/update-icons.js`, which checks for SVG files in the directory `src/icons`. It then uses [svgo](https://github.com/svg/svgo) to reduce the SVG image down to its simplest form, and then parses (using [htmlparser2](https://github.com/fb55/htmlparser2)) it to extract the path data from any `<path>` elements that it contains. (This might be an overly-naïve way to do this, but as we get more icons we can adjust it if it fails for any reason.)

This path data is stored in a map, which is written to `src/components/atoms/icon/icons.json`. This file is used as a lookup inside of `<Icon>`, using the `type` prop as the key.

I haven't finished the documentation piece yet (with searching, aliases, etc.), but I wanted to share this to get feedback on the approach. We can choose to merge this now, and then merge the documentation piece separately, or we can wait until both pieces are complete. I'm ok with it either way!

Let me know what you think!


nkohari:
I added documentation for `<Icon>` now. Here's a quick example of the `<IconBrowser>` component in action:

ezgif-4-55a2226e4c

(Ignore the fact that the icon is called `trash`, it's been fixed! 😄 )

Unfortunately, for some reason, I couldn't get the current version of `markdown-to-jsx` to render the `<IconBrowser>`, so I had to upgrade it to the latest version... which in turn broke the support for prop tables.

After some tinkering, I came up with a solution. Rather than the prop table being rendered for blocks beginning with `js props`, now it will render it for any block beginning with `jsx`. (Blocks marked as `js` are rendered as normal but the prop table is not included.) Hopefully this is alright; it was the best solution I could come up with.

siddharthkp:
Staging deployment failed because of build error: 

```
Module not found: Error: Can't resolve './icons.json' in '/home/nowuser/src/src/components/atoms/icon'
```

siddharthkp:
> After some tinkering, I came up with a solution. Rather than the prop table being rendered for blocks beginning with js props, now it will render it for any block beginning with jsx.

I don't mind at all `/shrug` It's all conventions that we are making up as we go.

nkohari:
Whoops, sorry about that! Forgot to add the `build:icons` step to the build process. This should be fixed now.

siddharthkp:
This is beautiful work 👌

Will back merge master and ship :shipit:

Good to have: How can we make copy paste easy after finding an icon?

siddharthkp:
Visual diff shows no regression 🌄👍

The sidebar does look funny though

before and after:

screen shot 2018-01-30 at 10 58 03 amscreen shot 2018-01-30 at 10 58 34 am

landitus:
The spacing issue might be because of the canvas. I see that also the new icons are misaligned vertically. I think we need to review what comes out of the SVG conversion or any processes that change the artwork in between.

nkohari:
You're right, let me make sure that I'm using the SVGs exported from the icon font, and then I'll tweak the padding in CSS to make it look like it should. I think the issue is that the SVGs go all the way to the edge of the canvas without leaving whitespace.

landitus:
It's looking really good!! Love the icon finder!

Regarding the API, I think it would be best to have a `size` enum with small, medium, large. Right now we are using a single size, right? So we can leave it as is and then review it. Regarding the color, it looks weird in the props table to have a token but it doesn't change the icon color unless its a hex value. I think we should set the default prop for color to #333. 

We could then merge this and test it further when using icons inside other components

nkohari:
1. Right now, `<Icon>` accepts a `size` prop which is a number, but we can change that to an enum if you want.
2. This is a bug with the props table in the docs; it should really try to resolve the value of that token rather than showing it literally. It happens with some other props too, but replacing the token with the literal `#ccc` will undermine the use of the token. :)
3. Agree with `name` instead of `type`. I actually made that mistake several times while building the component! I'll add a commit with that right now.

landitus:
Sounds good. About the size, I would leave it as is for now until we need multiple sizes.

Issue #187, opened 2018-01-24
Dial back on the design changes: take 3


landitus:
Change color of buttons to match current design. Primary color will be `lightBlue`.

PR #188, opened 2018-01-25
Changed color for primary button

landitus:
…as it is used today.

Details: https://github.com/auth0/cosmos/issues/187


siddharthkp:
Should we ship the `primaryAlt` button at all?

landitus:
I've updated the PR to only ship 1 primary button

siddharthkp:
tokens 💪

love how we were already able to do this before release 😄

PR #189, opened 2018-01-25
Implemented first version of Overlay and Dialog

nkohari:
This still needs some documentation, but I wanted to share these to get feedback.

screen shot 2018-01-25 at 1 30 45 pm

`<Overlay>` is an atom that provides a low-level interface to drawing something on top of the regular viewport. Right now a backdrop is always displayed, but we can also make that optional if we want to use this for things like `<Tooltip>` as well.

Here's an example of the Overlay API:

```jsx
<Overlay isOpen={this.props.isOverlayOpen} onClose={this.closeOverlay}>
  This content will be displayed in the overlay
</Overlay>
```

screen shot 2018-01-25 at 1 30 36 pm

The molecule `<Dialog>` then builds on `<Overlay>` to provide a higher-level component that looks like a dialog. It adds additional styling and props, and more structure including a titlebar, close button, and footer. It exports a couple of inner styled components, `<Dialog.Body>` and `<Dialog.Footer>`, which consumers should pass back as children.

Here's an example of the Dialog API:

```jsx
<Dialog title="Example Dialog" isOpen={this.props.isDialogOpen} onClose={this.closeDialog}>
  <Dialog.Body>
    The content to display in the dialog's body
  </Dialog.Body>
  <Dialog.Footer>
    <Button primary onClick={...}>OK</Button>
    <Button onClick={...}>Cancel</Button>
  </Dialog.Footer>
</Dialog>
```

You might notice that both components are managed, meaning that consuming components have to implement their own state management for the `isOpen` prop and a callback function that toggles this value, which is passed in via the `onClose` prop.

Let me know what you think! 😄


siddharthkp:
Feedback on the API:

1. Replace `isOpen` for `open`, only to follow the same convention as other components (Switch on, Tabs selected, etc.)

2. @landitus Can we standardise the footer? If you are only allowed to add actions to the footer, we can change the API to

```jsx
<Dialog.Actions
  primaryAction={{ label: 'Save Changes', method: this.save }}
  secondaryActions={[{ label: 'Clear', method: this.clear }]}
  destructiveActions={[{ label: 'Delete Client', method: this.delete }]}
/>
```

nkohari:
Sorry for the force-push (wanted to correct merge conflict w/o a merge commit). Thanks for catching the inconsistency with `isOpen`! I'm not sure I like the proposed change for `<Dialog.Actions>`... I understand the idea of locking down the `<Dialog>` but the API feels like it might be difficult to use.

siddharthkp:
> I'm not sure I like the proposed change for <Dialog.Actions>... I understand the idea of locking down the <Dialog> but the API feels like it might be difficult to use.

That's fair, have a look at [Form.Actions](https://auth0-cosmos.now.sh/docs/Form) as well. Happy to change the API for compound components to something better as long as it's predictable across

nkohari:
Just spitballing here, but what if we exported a function called `Dialog.action()` that could be used to create actions? This would help clean up the API, and we could even return an actual `DialogAction` object and use `PropTypes.instanceOf(DialogAction)`.

Something like:

```jsx
const MyDialog = props => {
  const actions = [
    Dialog.action("OK", this.onOkClicked, "primary"),
    Dialog.action("Cancel", this.onCancelClicked),
    Dialog.action("Delete", this.onDeleteClicked", "destructive")
  ]
  return (
    <Dialog actions={actions} ...>
      This is the body of the dialog
    </Dialog>
  )
}
```

The use of `Dialog.action()` might be confusing since it's pretty similar to creating a new component, so we could also just export a `DialogAction` class and let people do:

```jsx
const actions = [
  new DialogAction("OK", this.onOkClicked, "primary"),
  ...
]
```

WDYT? 🤔

siddharthkp:
This looks really powerful 💪 

I'm thinking of other places where utilities like this will help.

```jsx
Dialog.action("OK", this.onOkClicked, "primary")```

 

Does this affect documentation for prop types? Wondering if this wouldn't feel intuitive to folks it's not less React-y than 

```js
<Dialog actions={actions} ...>
```

siddharthkp:
P.S. Added an example of how to add images to docs: https://github.com/auth0/cosmos/pull/189/commits/7eed134cbb25d22cb4e55d0729a18b34fc4e734b

nkohari:
Really it's just an easy way to avoid having an implied interface for an object prop and trying to enforce it with `PropTypes.Shape()`. If we formalized actions using classes, like:

```jsx
class DialogAction {
  constructor(label, method, appearance = "default") {
    this.label = label;
    this.method = method;
    this.appearance = appearance;
  }
}
```

Then we can do:

```jsx
Dialog.propTypes = {
  actions: PropTypes.arrayOf(PropTypes.instanceOf(DialogAction)),
  ...
}
```

Then if we wanted to create a helper method so everything was hooked onto `Dialog`, we could, but it wouldn't be necessary:

```jsx
class Dialog extends React.Component { ... }
Dialog.action = (...args) => new DialogAction(...args);
```

This is really just my bias for TypeScript showing. 😄 

On a minor style note, IMO nesting arrays (and especially arrays of objects) inside of JSX always seems to become difficult to read. That's why I tend to go for local variables like `actions` in my example.

nkohari:
Is this ready to be merged, or did we want to change the API at all?

PR #191, opened 2018-01-29
Add Stack component

siddharthkp:
Component to horizontally layout children.

Trouble in aligning items:

If you let children take width according to their content (`flex: auto`), it lays out nicely.

screen shot 2018-01-29 at 6 19 03 am

&nbsp;

But, when the width is different across rows, it becomes inconsistent

screen shot 2018-01-29 at 6 18 51 am

&nbsp;

If you control the behaviour of the children (`flex: 1`), you get a more consistent look but a column might end up taking more space than it needs.

&nbsp;

screen shot 2018-01-29 at 6 18 31 am
equal width across rows ✅

&nbsp;

screen shot 2018-01-29 at 6 17 58 am
too much space for label ❌

&nbsp;

Update:

`Stack` needs to behave differently when used on it's own and when used in a list.

Here's are 2 possible ways to do it:

**Approach 1**: Change `flex` property of `StyledStack` from `List`.

We follow this pattern for some molecules like `ActionInput`, but the golden rule is: the parent can only change spacial arrangement of it's children, not the style of the children themselves.
Example here: https://github.com/auth0/cosmos/blob/master/src/components/molecules/_action-input/index.js#L16-L18

This solution feels dirty here because we want to change the style of a deeper nested child.

&nbsp;

**Approach 2 (the one I used here)**: Internally pass an extra prop when `Stack` is used in a `List` 

This makes more sense from a separation of concerns point of view, but the code to make it work isn't as pretty.

Here's the icky snippet:
```js
if (child.type.displayName === 'Stack') {
  return React.cloneElement(child, { consistentWidth: true })
}
```

If this block of code makes sense, then we can go ahead merge this :shipit: 

&nbsp;

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5002,15 +5002,6 @@
       "integrity": "sha1-2uRqnXj74lKSJYzB54CkHZXAN4I=",
       "dev": true
     },
-    "flexboxgrid2": {
-      "version": "7.1.0",
-      "resolved": "https://registry.npmjs.org/flexboxgrid2/-/flexboxgrid2-7.1.0.tgz",
-      "integrity":
-        "sha512-SQTLalHdPRHONA/UpJr9vEt4Nj8Aj6Bh/d+QES+MwzOS2rJ1MORHVbXbNlfnXQEnAuxp/DD7AimFb9DyZA5HOA==",
-      "requires": {
-        "normalize.css": "7.0.0"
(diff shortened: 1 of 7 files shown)


siddharthkp:
Merged, please review this along with `List` in https://github.com/auth0/cosmos/pull/197

Issue #174, opened 2018-01-11
Well Component


landitus:
There are 2 different visuals for forms. A single form is the default design, but when having multiple forms in a page, we have a different design (gray background and fieldset-type titles). This way, the multiple forms look distinct from each other. This pattern can be useful to use in the Advanced tab, where we are going to have multiple forms stacked.

Let's try a `Well` Component that can wrap a form to allow for this design.

Screenshot for reference:

<img width="889" alt="screen shot 2018-01-11 at 17 17 16" src="https://user-images.githubusercontent.com/239215/34845411-8b6924c2-f6f3-11e7-8d08-52ea7cfd711f.png">

PR #195, opened 2018-01-29
Well component and form adjustments

landitus:
Check out this idea to add a [Well component](https://github.com/auth0/cosmos/issues/174) as an easy way to layout multiple forms in pages. I don't think this is the best solution, but one that could work out for now. The idea is to use a well to add a visual separation to forms. Using existing components like fieldsets, you can create a title to the form. This is all composable and looks ok. 

At the same time, I've adjusted the spacing of **form fields**, **form actions**, **fieldsets** and **buttons** so that stacking these elements is more predictable. It was all sort of connected to this task so I did it on the same PR.  

As a bonus, I've added the Advanced tab and filled it with our current set of components... and it's looking pretty nice!

screen shot 2018-01-29 at 18 30 57


siddharthkp:
Looks amazing!

The only thing I'd change is the API. We should make `Well` an internal component, and create a new compound can called `FormGroup` that wraps it's children in wells

src/components/atoms/well/index.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+
+import { colors, misc, spacing } from '../../../tokens/'
+
+const StyledWell = styled.div`
+  background: rgb(249, 249, 249);
```
siddharthkp: tokensssss

src/components/molecules/form/divider.js
```diff
@@ -1,17 +1,15 @@
...
 import { fonts, spacing, colors } from '../../../tokens'
-import StyledLabel from './label'
 
-const StyledDivider = styled(StyledLabel)`
-  text-transform: uppercase;
+const StyledDivider = styled.div`
   text-align: left;
-  font-size: 13px;
-  letter-spacing: 1px;
+  display: block;
```
siddharthkp: probably redundant line. `div` are block by default

src/manage/advanced.js
```diff
@@ -15,7 +15,243 @@ class Advanced extends React.Component {
...
+          </Form>
+        </Well>
+        <Well>
+          <Form>
+            <Form.FieldSet label="Certificates">
+              <Form.TextArea
+                label="Signing Certificate"
+                code
+                readOnly
+                value="-----BEGIN CERTIFICATE-----
```
siddharthkp: Where did this certificate come from? The accuracy of the format worries me 😅

src/components/atoms/well/index.js
landitus: got me!

src/components/molecules/form/divider.js
landitus: agree

src/manage/advanced.js
landitus: I have my own testing account, where I've changed some random letters, but probably not enough. I'll do some more faking so that it doesn't look so real.

landitus:
> The only thing I'd change is the API. We should make Well an internal component, and create a new compound can called FormGroup that wraps it's children in wells

Would the code look like this?

```
<Form.Group>
     <Form>
           ....
      <Form>
</Form.Group>
<Form.Group>
      <Form>
           ....
      <Form>
</Form.Group>
```

siddharthkp:
I imagine this to be higher up than Form, so it can't be Form.Group

How about FormGroup?

landitus:
I've done the Form.group approach, but now I'm going to try the FormGroup approach as you suggested

landitus:
Here's the FormGroup approach! Now that I look at it, it looks kinda like a Well component, duplicated. The Form.Group was just one line of code (and one example in the Form docs) and this approach is a whole new component, but one very simple. wdyt?

siddharthkp:
> Here's the FormGroup approach! Now that I look at it, it looks kinda like a Well component, duplicated.

I think that's because the API needs a tiny touch up.

You should only have to mention `FormGroup` once. After all, it's a group with multiple forms in it.

```jsx
<Form.Group>
  <Form>....</Form>

  <Form>....</Form>
</Form.Group>
```

The group should take care of wrapping `Well` on each + space between consecutive forms.

It's still a tiny component, but I imagine it to look something like this:

```jsx

const StyledFormWrapper = styled(Well)`
  margin/padding
`

const FormGroup = props => {
  return React.Children.map(props.children, child => <StyledFormWrapper>{child}</StyledFormWrapper>)
}

FormGroup.propTypes = {
  children: PropTypes.arrayOf(PropTypes.instanceOf(Form))
}
```

siddharthkp:
Sidenote: There are some lint warnings on the console (unused variable), not a big deal, but everyone likes a clean console 😄

landitus:
Will check those as well! thanks for the heads up

landitus:
Something weird happens:
- I can't seem to get the <StyledFormWrapper> styled here. Any clues?
- When I add the loop code, the doc page disappears and the component appears undocumented.

So far I'm stuck!

siddharthkp:
Push the broken code, I'll have a look 👍

siddharthkp:
My bad, my loop code doesn't add a container for the children. 

So the error is this one: `Adjacent JSX elements must be wrapped in an enclosing tag`

That error was hidden behind multiple other things that it broke 😅

landitus:
Got the styles to work, thanks! Still, the docs for FormGroup are behaving badly. When you visit the link in the docs, nothing happens (and I didn't see any console errors)

PR #197, opened 2018-01-30
Client List

siddharthkp:
- [x] Cleaned up API for List and Stack
- [x] Mocked up client list


siddharthkp:
@landitus Up for review.

Check out the docs for Stack and List, I'm not super happy with the API

siddharthkp:
On second thoughts `List.Item` is a horrible idea. Going to scrap it.

I'm trying too hard to make the list a table. It's not a table.

landitus:
It's looking good, I think we only need to tweak some CSS for the mocked-up Clients list. But I like the components. Should we merge this and then do the CSS tweaks?

landitus:
This is something I've realized: How can we align these buttons to the right? 

screen shot 2018-02-06 at 01 52 22


I think each element in the stack would need to be aligned either left, right o center. This a behaviour which happens in tables (or table-like layouts) all the time. wdyt?

siddharthkp:
> It's looking good, I think we only need to tweak some CSS for the mocked-up Clients list. But I like the components. Should we merge this and then do the CSS tweaks?

Yep, let's do in a different pull request. It's not part of `Stack`

> We should discuss how to achieve different alignments in each column of the Client list

Removed compressed and added `align` instead. This should take care of it.

PR #201, opened 2018-01-31
Page header

beneliflo:
Well after many changes and see how to accommodate the components this is what I have now.
I'm still not satisfied because there are many things that I think are missing.

1. Is there not an option to change the color of a button? Clients has the orange button.
2. Do we need to create a component for avatars / images?
3. For the icons that are not in budicon, how should we proceed?

Some styles are still missing but the most important thing is how you see the structure and what you think about that.

PS: Is `Page Header` or `Pager Header`?


beneliflo:
Rebase with master

landitus:
Thanks Benja! 

Let me clarify:

> Is there not an option to change the color of a button? Clients has the orange button.
Lets use our new primary button, which is Light blue

> Do we need to create a component for avatars / images?
I think we do. For now, just leave it hardcoded and then we'll extract it as a component

> For the icons that are not in budicon, how should we proceed?
Good question! @nkohari we need to think about this. Maybe we can let users pass an svg or something?

> Some styles are still missing but the most important thing is how you see the structure and what you think about that.
Yes, that's a good idea, we can always retouch the styles later. I think it would be a good idea to separate this component: the default Page header will be just Page header and live inside the molecules. This will be used 95% of times. The special Clients Header can live inside manage as it's more custom and it's only used one time. It's more of an example of a local component.

> PS: Is Page Header or Pager Header?
It's Page header

landitus:
wdyt @siddharthkp ?

siddharthkp:
@beneliflo Can you resolve the merge conflict? Travis doesn't run it's thing 😢

src/components/index.js
```diff
@@ -68,3 +71,6 @@ export { Stack }
 
 import List from './molecules/list'
 export { List }
+
+import PagerHeader from './molecules/pager-header'
```
siddharthkp: `PageHeader`

src/manage/client-list.js
```diff
@@ -1,19 +1,23 @@
...
       <div>
-        <Heading1>Clients</Heading1>
-        <div>client list</div>
-        <br />
-        <br />
+        <PagerHeader
+          title="Clients"
+          description={{
+            text: 'Setup a mobile, web or IoT application to use Auth0 for Authentication.',
+            learnMore: '/clients'
```
siddharthkp: @landitus Can we freeze this in the API?

I like how this looks, but not sure if it fits all the use cases

src/manage/client-list.js
```diff
@@ -1,19 +1,23 @@
...
+          }}
+          primaryAction={{ label: 'Create Client', icon: 'plus', method: this.save }}
+          secondaryAction={{ label: 'Turorial', icon: 'video', method: this.save }}
```
siddharthkp: 👍

siddharthkp:
Structure looks good to me!

The only doubt is the description prop

```js
description={{
  text: 'Setup a mobile, web or IoT application to use Auth0 for Authentication.',
  learnMore: '/clients'
}}

I like how it creates a convention. @landitus, can we get away with this? Or does it fail in some use cases?

@beneliflo docs page please! That's the best way to review a component

File: src/components/atoms/breadcrumb/index.js [hunk 4]
```diff
@@ -0,0 +1,21 @@
...
+
+import { colors, misc } from '../../../tokens/'
+
+const StyledBreadcrumb = styled.a`
+  color: rgba(0, 0, 0, 0.56);
+  font-size: 13px;
+  text-decoration: none;
+`
+
+const Breadcrumb = ({ href, content, ...props }) => {
```
siddharthkp: Scope creep? 😉

src/manage/client-list.js
landitus: Yep, I think we're good 👍

beneliflo:
@landitus @siddharthkp How do you see it now?

I just separated into `top` when using` breadcrumb` or `actions` (buttons). Since one or the other will be used only once in each case.

And the differents both page header is in molecule component. How do you see that?

Beginning to document...

src/components/molecules/page-header/index.js
```diff
@@ -0,0 +1,71 @@
...
+    Top = <Breadcrumb content={props.breadcrumb.content} link={props.breadcrumb.link} />
+  } else {
+    return null
+  }
+
+  return (
+    <StyledPageHeader>
+      {Top}
+
+      <div>
```
landitus: In the case of the Client header (the complicated one), I think we could reorder some elements here to make it work with flexbox. I think that flexbox makes it easier to compose elements. We could organize the component like this:

```
<div> <<-- With flexbox
   <Logo /> 
  <div>
    <Heading>
    TYPE ... ID etc
  </div>
<div>
```

src/manage/client-list.js
```diff
@@ -26,14 +33,23 @@ class ClientList extends React.Component {
...
-        <br />
+        <PageHeader
+          title="Clients"
+          description={{
+            text: 'Setup a mobile, web or IoT application to use Auth0 for Authentication.',
+            learnMore: '/clients'
+          }}
+          actions={{
+            primaryAction: { label: 'Create Client', icon: 'plus', method: this.save },
+            secondaryAction: { label: 'Turorial', icon: 'video', method: this.save }
```
landitus: Check the typo in `Tutorial`

landitus:
> Add docs please

Yep! The docs are the best way to test this component. Can you add the docs? You can base them from the _box.md example file.

src/components/molecules/page-header/index.js
landitus: wdty?

PR #202, opened 2018-02-01
Document Form children by moving them into their own files

siddharthkp:
- [x] Move all children to their components
- [x] Add documentation for children
- [x] Change tooling + docs to accommodate this

Todo in a different PR:
- Make strings work in prop table


siddharthkp:
@landitus Check out the docs on staging. Have pulled out and documented `Form.Actions`

landitus:
Looks better!

siddharthkp:
> This also calls for improving, in time, the props table and maybe we can think of a way of better formatting the array, so that it doesn't look like raw code.

The types part? Isn't that supposed to be raw code?

landitus:
Ok! I've seen some docs that show the word array and only the keys. So it's just an idea at this point.

siddharthkp:
> I've seen some docs that show the word array and only the keys. So it's just an idea at this point. If it doesn't make sense, ignore it :)

Gotta show the type along with key /shrug

Issue #185, opened 2018-01-23
Find a better input for uneditable tokens


landitus:
![input-password-tokens](https://user-images.githubusercontent.com/239215/35281614-0fb478a2-0032-11e8-9916-b4a4dd12206e.gif)

The [input component](https://github.com/auth0/cosmos/issues/125) (for passwords) triggers 1Password and other login related features when saving a form with the token fields. We should find a better and more semantic field for displaying tokens. Using another component (not a field) is an option.

landitus:
This plain text component seems a good solution to replace the input. It would be useful to show uneditable text and also for tokens. We would need a hidden prop to mask the contents of the token. Making it part of the inputs family allows to match the alignment with the labels and with the rest of the fields, though visually it doesn't look like an input. 

<img width="1098" alt="screen shot 2018-01-23 at 11 26 14" src="https://user-images.githubusercontent.com/239215/35280898-54e5f0f6-0030-11e8-8770-948bef81d553.png">

landitus:
Looking at what we already have, we could get away with using an input field (read-only) and have a special prop that masks the value with the circle character. That prop will allow to reveal the token and also copy it with the icon button.

siddharthkp:
I don't understand the problem fully. 

Is this because there is a `type=password`? Are we missing a html prop that stops this behavior?

landitus:
Yes, because this is a type=password, 1Password and Google Smart Lock will trigger once you save the form. I've done some research and there's no special property to disallow this. Looking at what Manage uses today, I came up with this proposal, that avoids using a password field to mask data. 

I think that there are other solutions on the table, that could be done later, like **not** using inputs when having uneditable or masked data. Just using a Text component that changes to an input when you want to edit. I've seen this UX pattern many times.

I like the idea of a better UX for handling tokens and sensible data as this is core to Auth0.

hzalaz:
The problem of the password mangers is real since everytime we edit a pwd field (for whatever reason) and submit a form we are prompted to save the password 😄 . It can be a bit annoying since you could be working with an RSA key or a HS secret or event creating a user via manage and prompting a password (or even trying a DB connection to see if it works).

The solution of the read only will help us with those attributes that cannot be edited but in some cases we also need to edit some of them

nkohari:
Closing since since we now have masked input, even if we may want to improve it. :)

PR #205, opened 2018-02-01
[TextInput] Add new "masked" prop to obscure hidden data

nkohari:
As described in #185, the use of `type=password` to hide non-password information triggers password managers like 1Password. This patch adds a prop called `masked` to the `TextInput` which will manually obscure the value, and force the input to be `readOnly`.

If we decide to go with this proposal, we should definitely test switching a managed `TextInput` between masked and "not-masked" to make sure the value isn't lost. If the component is managed (that is, the`value` comes from a parent component) it should be okay.

WDYT?

### DIFF
--- a/src/components/atoms/text-input/index.js
+++ b/src/components/atoms/text-input/index.js
@@ -9,10 +9,12 @@ const StyledInputElement = StyledInput.extend`
 
 const TextInput = props => {
   if (props.masked) {
-    const length =
-      props.defaultValue && props.defaultValue.length > 0 ? props.defaultValue.length : 8
+    const { value } = props
+    const length = value && value.length > 0 ? value.length : 8
     const maskedValue = new Array(length).join('•')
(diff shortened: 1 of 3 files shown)


landitus:
I like the idea! In the docs I only see the field turned into a read-only, but the value is still visible. Maybe it's a matter of replacing the characters with the circle symbol to mimic the password field output: https://stackoverflow.com/a/37965943/272553

nkohari:
I changed this slightly to use the `placeholder` property instead of `defaultValue`, and to null out any value that would normally be passed.

This means that the value of the `<input>` will actually be `null` if you read it from the DOM element itself, but this should be okay as long as the component isn't intended to support "regular" HTML forms. (If the `<form>` the `<TextInput>` is in is submitted, the form body won't include the value for masked `<TextInput>`s, because they've been set to null.)

I also added an example in the docs to illustrate how this works. @siddharthkp, WDYT about this?

PR #211, opened 2018-02-06
Docs styles

landitus:
Tweaked the styles of the Documentation to allow for a more predictable markdown output. It's mostly typography. To prevent collisions, I've added local components to the Docs.

Now we can use `H2` (##) for `Section titles` and `H3` (###) for `Example titles` and they all render with consistent spacing and separation. Also tweaked the Sidebar (and the Logo component). The docs will look more polished!


landitus:
I was trying to add a "description" to the Frontmatter, to be able to style it differently, but wasn't able to render it in the docs. I added an example of the description in the Empty State component.

PR #213, opened 2018-02-06
Added separation for buttons

landitus:
Now that buttons don't have margin or spacings, I thought to add a hidden `Actions` component that allows to wrap a set of buttons or actions in an inline list. At first I thought this could be done with the `Stack` component but the result is not the same (there are paddings and the buttons are affected by the width. Using this `Actions` component, it adds a list with the list items as inline-block. This is very predictable and works pretty well for Form.Actions.

wdyt?

### DIFF
--- a/src/components/atoms/_actions/index.js
+++ b/src/components/atoms/_actions/index.js
@@ -1,0 +1,33 @@
+import React from 'react'
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+
+import { spacing } from '../../../tokens/'
+
+const StyledActions = styled.ul`
+  list-style: none;
+  margin: 0;
+  padding: 0;
+`
+
+const StyledItem = styled.li`
+  display: inline-block;
+  margin-right: ${spacing.xsmall};
+  &:last-child {
+    margin-right: 0;
+  }
+`
+
+const Actions = props => {
+  const children = React.Children.map(props.children, child => {
+    return <StyledItem>{child}</StyledItem>
+  })
+
+  return <StyledActions>{children}</StyledActions>
+}
+
+Actions.propTypes = {}
+
+Actions.defaultProps = {}
+
+export default Actions
--- a/src/components/molecules/form/index.js
+++ b/src/components/molecules/form/index.js
@@ -11,7 +11,8 @@ import TextArea, { StyledTextArea } from '../../atoms/textarea'
 import Select from '../../atoms/select'
 import Switch from '../../atoms/switch'
 import Button from '../../atoms/button'
-import { Right, Clear } from '../../_helpers/float'
+import Actions from '../../atoms/_actions'
+import { Right, Left, Clear } from '../../_helpers/float'
 
 import StyledLabel from './label'
 import StyledDivider from './divider'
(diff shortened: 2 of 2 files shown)


siddharthkp:
Is _actions the same as a button group component?

landitus:
I think they are pretty similar. Actions in my mind has space between buttons, and the button group (as we have) does not. But this is subjective and we can see what fits best.

See this examples: https://getbootstrap.com/docs/4.0/components/button-group/ and https://polaris.shopify.com/components/actions/button-group. In UI, people tend to refer to button group when actions are sticked together (like bootstrap) in a toolbar, like Gdocs or Github comments section. Sometimes they even double for checkboxes and radios.

We don't have buttons sticked together (yet). So we could use the button actions components with spacing. Visually and UX wise, we want to achieve a component (like Polaris) that allows:

>  to group displays multiple related actions in a row to help with horizontal arrangement and the spacing of calls to action.

siddharthkp:
From the components we have, does the component styles make sense only in the context of Form Actions?

siddharthkp:
> Actions in my mind has space between buttons, and the Button group (as we have) does not

We can decide that 😄 , can also depend on it's context

landitus:
The actions / button groups (inline buttons with separation) are used right now in the Form Actions (even Modal actions), Client (or any) List/Table row actions and in the Input action component

landitus:
This is the **Client (or any) List/Table row actions** example. We'll need to add proper separation with the actions or button group component.

screen shot 2018-02-06 at 11 24 57

**Form actions/modal actions**

**Input actions:**

screen shot 2018-02-06 at 11 26 30

siddharthkp:
Can we call it ButtonGroup and use it everywhere? I'm scared of the idea of 2 components doing kind of similar things

landitus:
Sounds good to me!

landitus:
Cool, I'll give it a shot

landitus:
I was looking at the Stack more in detail, maybe it's slightly off-topic, but I found the default aligning odd in terms that it distributes the elements to fill the available space. I think that behavior is very unpredictable. Even in table like situations, different lengths of content will not be aligned and create the illusion of rows. I think we need to review what to do with the Stack, the ButtonGroup and the List.

siddharthkp:
That's fair. The more I use it, the more I feel we just need to build a table.

landitus:
In terms of CSS I think we can fix the List, but we need to part ways with the Stack for it. What would be the differences between the List and the Table?

We could look at some examples: 
- https://ant.design/components/list/#components-list-demo-loadmore
- https://polaris.shopify.com/components/lists/resource-list#navigation

Or maybe, as you suggest, we should go full on Data Table like:
- https://ak-mk-2-prod.netlify.com/packages/elements/dynamic-table

siddharthkp:
> What would be the differences between the List and the Table?

List doesn't control columns, only rows.

= 

In the case of List, the responsibility of consistency between rows is on the developer.

siddharthkp:
@landitus Another thing to consider here: Button width in https://github.com/auth0/cosmos/pull/264

siddharthkp:
Implemented in https://github.com/auth0/cosmos/pull/284

Issue #216, opened 2018-02-06
Extract meta tags from docs 


siddharthkp:
```
category: Form
description: something something
```

and add it to `metadata.json`

PR #222, opened 2018-02-07
Add doc page-header component

siddharthkp:
You can drop information at the top of the doc using [yaml](http://docs.ansible.com/ansible/latest/YAMLSyntax.html) like this: 

```meta
  category: Actions
  description: Buttons are great to ask users for action
```

and it will be available in the doc's [`PageHeader`](https://github.com/auth0/cosmos/compare/doc-header?expand=1#diff-ce2b0d4526663a22e8690c07c90193a5) component as props.

Check out  Button [markdown](https://github.com/auth0/cosmos/pull/222/files#diff-0497367b9a8e0c9306ebdd7854d050e5) and [docs](https://auth0-cosmos-doc-header.now.sh/docs/Button) to see an example.

The language choice is very arbitrary. I saw someone use yaml in markdown, so I started using it, we can also use json. That would look like this:

```meta
  {
    "category": "Actions",
    "description": "Buttons are great to ask users for action"
  }
```

Fixes https://github.com/auth0/cosmos/issues/216

### DIFF
--- a/src/docs/spec/page-header.js
+++ b/src/docs/spec/page-header.js
@@ -1,0 +1,12 @@
+import React from 'react'
+import styled from 'styled-components'
+
+import { Subheader } from '../docs-components/typography'
+
+const Header = props => (
+  <div>
+    <Subheader>{props.description}</Subheader>
+  </div>
+)
+
+export default Header
--- a/src/docs/spec/page-header.js
+++ b/src/docs/spec/page-header.js
@@ -1,12 +1,17 @@
 import React from 'react'
 import styled from 'styled-components'
 
-import { Subheader } from '../docs-components/typography'
+import { Heading1, Subheader } from '../docs-components/typography'
+
+const Headings = styled.div`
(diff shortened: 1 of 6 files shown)


landitus:
IMO it's ok to go by convention here! In this case, this is a frontmatter section of a markdown file and I guess everyone uses yaml in this case. https://jekyllrb.com/docs/frontmatter/

src/docs/spec/page-header.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import styled from 'styled-components'
+
+import { Subheader } from '../docs-components/typography'
+
+const Header = props => (
+  <div>
+    <Subheader>{props.description}</Subheader>
+  </div>
```
landitus: Can we place the description in a `<Subheader>` like in the previous version of this file? I meant for it go here 😉

siddharthkp:
Fixed!

PR #223, opened 2018-02-07
Fix route conflict for docs

siddharthkp:
now folks released a new feature, we got the feature because we don't lock our version of now environment. Feature/Bug/whatever.

The `public` folder is treated as a public route for sharing assets. This layer is before our react router layer.

`public/docs` got preference over our app's `/docs` route.

Fix: I renamed `public/docs` to `public/screenshots` because that's what we're using it for 😂

Will raise an issue for long term fix, can't be bothered before the PoC 😛


siddharthkp:
merge merge merge?

PR #225, opened 2018-02-07
Better documentation for Tabs

landitus:
- [x] Added Playground
- [x] Document `selected` prop (manually)

I got stuck documenting the `selected` prop. I added the `boolean` type but it looks weird as the switch doesn't do anything as it can't know which tab to activate. That made me look into other implementations where the selected tab is a `string` passed to **Tabs** (not to the single Tab) and each Tab has an `ID`. Just throwing this to see if we can solve this now, or we sit on this.

Check the deployment docs look like this:
screen shot 2018-02-07 at 20 28 51

### DIFF
--- a/src/components/molecules/tabs/index.js
+++ b/src/components/molecules/tabs/index.js
@@ -69,8 +69,7 @@ Tabs.Tab = props => <TabContent>{props.children}</TabContent>
 
 Tabs.propTypes = {
   /** Children should be an array of React elements */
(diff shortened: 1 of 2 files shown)


siddharthkp:
`selected` is a property of `Tabs.Tab`, not `Tabs`

landitus:
How would that look? I did a quick try and It looks a little weird:

screen shot 2018-02-08 at 11 27 57

PR #226, opened 2018-02-08
Switch alignment and style adjustments

landitus:
- [x] ⭐️💅Fix vertical alignment of witch with label in Form.Field   
- [x] 💅Fix text style for Label to use the proper uppercase style


src/components/atoms/switch/index.js
```diff
@@ -4,6 +4,11 @@ import PropTypes from 'prop-types'
 
 import { colors, spacing, misc } from '../../../tokens/'
 
+const StyledSwitch = styled.span`
+  display: inline-block;
+  vertical-align: middle;
+  height: 32px;
```
siddharthkp: Should this height come from a common place?

src/components/atoms/switch/index.js
```diff
@@ -52,6 +57,9 @@ const Toggle = styled.span`
 const Label = styled.label`
   vertical-align: top;
   line-height: 32px;
+  font-size: 13px;
```
siddharthkp: tokenssss

src/components/molecules/form/field/index.js
```diff
@@ -28,6 +29,9 @@ const StyledField = styled.div`
     min-height: 44px;
     /* TO-DO: Tokenize this value as it's related to the other heights of components. */
   }
+  ${StyledSwitch} {
+    margin-top: 6px;
```
siddharthkp: This is a non standard value 😢

landitus:
It could. I had to declare the height to get it to align. In this case the height is part of the scale, but I don’t think we should use a spacing token here.

landitus:
> We can add a todo to tokenise them later

I agree.

> There's a 6px margin which is a non standard value 😢, can we make it 8?

that 6px corresponds (44-32)/2 so I could express it an operation between variables.

siddharthkp:
> that 6px corresponds (44-32)/2 so I could express it an operation between variables

umm, that's seems forced. Can we get away with away with `margin-top: 8px;`? If it's internal to the component, we can have non-standard values. I think we have some of those to solve alignment issues with select/textarea

landitus:
Yes, this is that case where it's just internal for this component. Any other value makes the switch look off center.

src/components/molecules/form/field/index.js
landitus: I had to make this adjustment when the switch is in the Field (left-aligned) it needs this notch of margin.

landitus: I think we need to review this in the top-align version. I'll add a big form in the overview to test all fields with top alignment.

siddharthkp:
I think you just need to tokenise font size and we're good!

landitus:
awesome! will do

landitus:
I've added the tokens, and also fixed a small spacing that the hidden `<input>` before the Switch was creating. The solution is to remove the element from the DOM flow. I've avoided using `display: none` because maybe it affects the form or accessibility. Did a position absolute instead. But I'm open to ideas of the best way to solve this.

Issue #253, opened 2018-02-09
Evaluate variables in prop types


siddharthkp:
TLDR: To make icon names a dropdown instead of string in the docs, I need to invest a day in some configuration crap. Need help to decide priority.

--

Long story:

This one took me by surprise!

Working on https://github.com/auth0/cosmos/issues/231, I landed on docs for [`Icons`](https://auth0-cosmos.now.sh/docs/Icon), this is a beautiful component. 

@nkohari Sorry if you asked about this earlier and I missed the conversation.

Here are the prop types for Icon:

```js
Icon.propTypes = {
  /** Icon name */
  name: PropTypes.string.isRequired,
  /** Icon size */
  size: PropTypes.number,
  /** Icon foreground color */
  color: PropTypes.string
}
```

Now that we have `enum` support in the docs, I wanted `name` to be a dropdown from the list of icons we have.

So I changed the prop-types definitions

```diff
+ const iconNames =  Object.keys(icons)

Icon.propTypes = {
  /** Icon name */
-  name: PropTypes.string.isRequired,
+  name: PropTypes.oneOf(iconNames).isRequired,
  /** Icon size */
  size: PropTypes.number,
  /** Icon foreground color */
  color: PropTypes.string
}
```

Turns out, `react-docgen` does not evaluate variables inside propTypes, which makes the prop table useless

```js
{
  name: "enum",
  value: "iconNames"
}
```

A nice and clean way to fix this is to cause evaluation the value of `iconNames`. [`preval`](https://github.com/kentcdodds/babel-plugin-preval) is a nice way of doing it, but it requires modifying the babel config. As we use `create-react-app`, we don't have access to that configuration 🙃

We can eject (sad) or use something like [react-app-rewired](https://github.com/timarney/react-app-rewired) to gain control of the config. 

OR we can upgrade to `react-scripts-next` (experimental) to use `babel-macros` to do the same 🎉 

I can invest some time in getting this all hooked up, but not sure what's the priority of this, if the ROI is big enough, I can probably pull this off in a day. @nkohari, need your help here!

Related: `color` should probably come from our tokens instead of `Proptypes.string`? @landitus, wdyt?

landitus:
My take: we could do it after the PoC.
- I don't mind having a string for icons, at least for now. Having a dropdown of icons is a good enhancement to the docs.
- Just to take into account: Should we need support for custom icons in the future; maybe not passing a name and passing a svg directly?
[...]

PR #257, opened 2018-02-09
Proptable clean strings

siddharthkp:
**Hard dependency on https://github.com/auth0/cosmos/pull/255** (Review after merging that)

Fixes https://github.com/auth0/cosmos/issues/231

&nbsp;

Part 1:

Because `react-docgen` does not evaluate `propTypes` and `defaultProps`, it results in ugly strings in the metadata

```js
name: "'copy'"
```

Added some string cleaning in `build:docs`. (Not the prettiest code, we can rid of this part after https://github.com/auth0/cosmos/issues/253)

&nbsp;

Part 2: 

Cleaned up potential cases for prop-string generation for playground. The code is now readable 🎉 and handles booleans, numbers and strings cleanly.

This is how it works:

```
Case 1: Falsy booleans:  These should be ignored: primary="false" / name="null"

Case 2: Truthy boolean: We only need to drop the key
Example: primary="true" results in <Button primary>

Case 3: Strings: They should be printed wrapped in double quotes "string"
Example: <Button icon="copy">

Case 4: Numbers: numbers are wrapped in in {}
Example: <Icon size={20}>

Default case: If something reaches here, we probably don't know what to do with yet. Wrapping the value in {} is a safe output
```

### DIFF
--- a/tooling/component-metadata.js
+++ b/tooling/component-metadata.js
@@ -30,6 +30,25 @@ const run = () => {
         /* parse the component code to get metadata */
         const data = docgen.parse(code, null, handlers)
 
+        /* remove redundant quotes from default value of type string */
+        Object.values(data.props).forEach(prop => {
+          if (prop.defaultValue) {
+            if (typeof prop.defaultValue.value === 'string') {
+              prop.defaultValue.value = prop.defaultValue.value.replace(/^'(.*)'$/, '$1')
+            }
+          }
+        })
+
+        /* remove redundant quotes from enum values in prop types */
+
+        Object.values(data.props).forEach(prop => {
+          if (prop.type.name === 'enum') {
+            prop.type.value.forEach(element => {
+              element.value = element.value.replace(/^'(.*)'$/, '$1')
+            })
+          }
+        })
+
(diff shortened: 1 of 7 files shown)


tooling/component-metadata.js
```diff
@@ -30,6 +30,25 @@ const run = () => {
         /* parse the component code to get metadata */
         const data = docgen.parse(code, null, handlers)
 
+        /* remove redundant quotes from default value of type string */
+        Object.values(data.props).forEach(prop => {
+          if (prop.defaultValue) {
+            if (typeof prop.defaultValue.value === 'string') {
+              prop.defaultValue.value = prop.defaultValue.value.replace(/^'(.*)'$/, '$1')
```
nkohari: Do we also need to support double quotes? I'm not sure if `prettier` takes care of this or not, but maybe:

```
/^['"](.*)['"]$/
```

siddharthkp: Yep, our precommit hook gets rid of double quotes

nkohari: Cool! I'll merge this as-is then.

Issue #250, opened 2018-02-08
[Paragraph] Should have a margin-bottom, as with an HTML <p> element


nkohari:

siddharthkp:
How much though?

Issue #248, opened 2018-02-08
[Paragraph] Add documentation


nkohari:

siddharthkp:
We have a `Paragraph` component? 🤔

landitus:
Actually we have a `Text` Component. But it's a span and I think It would be more useful if it's a <p> element. The main difference with a p (and styling) is that when you have multiple lines of text, they get separated. Also, being a block element, is easier to compose. 

Question: would it be ok to just use a `p` tag? or we use `Text` or `Paragraph` ?

nkohari:
Is it possible/acceptable to apply the basic typographic rules (font family, font size) globally? If so, we don't need a `<Text>` tag. This could also allow us to eliminate all of the typographical components like `<Paragraph>` and the `<Heading>`s as well.

Keeping headings might be useful in case we want to apply other rules to them; for example, we may want all headings to have an anchor associated with them to enable deep-linking/tables of contents, etc. I'm not sure where we want to draw the line on components that just add styles.

To be honest, `<Link>` could fall into this category as well, but I can see the value in keeping it around if we want to add support for client-side routing.

siddharthkp:
We do already some font stuff globally: https://github.com/auth0/cosmos/blob/master/src/components/atoms/typography/index.js#L5-L14

Can add `p`, `section`, etc. to that.

I'm not sure if we can add margin for `p` or font size for headings globally because of the way folks will transition to cosmos from their apps. We might end up overwriting their styles, our components should be opt-in

landitus:
Makes sense. In that case we should go for the component route for the text, headings, links elements as well, right?

nkohari:
Another option to throw into the mix: we could add a common CSS class like `cosmos` to all of our components, which would allow us to apply the styles globally across all components in Cosmos without leaking the styles. For example:

```
.cosmos {
  font-family: ...;
}
.cosmos p {
  margin-bottom: ...;
}
```

The more I consider it, the more I think we might want to avoid components for typographical elements unless it's necessary. It may just be adding more things that users need to learn without bringing a whole lot of new functionality to the table.

Issue #120, opened 2018-01-04
Text Component


nkohari:
<img width="702" alt="screen shot 2018-01-04 at 16 53 02" src="https://user-images.githubusercontent.com/1576/34581704-745744ec-f15f-11e7-8e0d-1bc79b402634.png">

landitus:
Should we call it Paragraph?

nkohari:
We may want both a `<Text>` and a `<TextBlock>` (or `<Paragraph>`) component. The former representing a styled `<span>`, and the latter representing a styled `<p>`.

PR #268, opened 2018-02-13
Document header + paragraph

siddharthkp:
Header

Have done a massive refactor of the API, now it's much cleaner to use.

- [x] Removed `Header1`, `Header2` etc. in favor of `Header size={n}`
- [x] Renamed `Subheader` to `Heading size={5}`
- [x] Updated internal components that used them
- [x] Updated usage in PoC 

screen shot 2018-02-14 at 12 15 17 pm

Paragraph: 
- [x] Fixes https://github.com/auth0/cosmos/issues/120 
- [x] Fixes https://github.com/auth0/cosmos/issues/249
- [x] Fixes https://github.com/auth0/cosmos/issues/248
- [x] Fixes https://github.com/auth0/cosmos/issues/250 (adding `{spacing.small}`)

### DIFF
--- a/src/components/_helpers/globals.js
+++ b/src/components/_helpers/globals.js
@@ -1,6 +1,6 @@
 import { injectGlobal } from 'styled-components'
 
-import { fonts, misc } from '../../../tokens'
+import { fonts, misc } from '../../tokens'
 
 injectGlobal`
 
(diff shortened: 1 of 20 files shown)


siddharthkp:
Need help on the API here:

`Header size={1}` means `h1`
`Header size={5}` means `h5`

Intuitively, this seems wrong. size 5 should be bigger than size 1.

PR #269, opened 2018-02-14
Add conflict handling for prop table

siddharthkp:
🎉

switches


landitus:
Awesome!

landitus:
I can't see the deployment, but locally it looks good. Just catched a bug thanks to this. :) When the buttons are disabled, most combined states stay the same. For example:

screen shot 2018-02-14 at 18 12 42

The disabled destructive button should be visually dimmed.

landitus:
But this LGTM!

siddharthkp:
Fixed in https://github.com/auth0/cosmos/pull/280

PR #273, opened 2018-02-14
Set font size of code input/textareas to 13px, force fixed height, remove white-space pre in textareas

### LINKED ISSUE #243: [TextArea] Long lines trigger horizontal scrolling, but text should wrap to the next line instead
Here's an example from the Advanced tab on the Clients screen in the PoC:

screen shot 2018-02-08 at 12 16 38 pm

### LINKED ISSUE #244: [TextArea] Font size of code should be smaller than regular text
Should be 12/13px?

nkohari:
Closes #243 and #244.

We may want to consider adding tokens for the font size and/or the input height, but this should suffice for now.


siddharthkp:
LGTM!

landitus:
@nkohari I think the **forced height** isn't in master anymore!

PR #280, opened 2018-02-15
Fix disabled state for Button

siddharthkp:
disabled is a state, not a appearance.

`disabled` can be used along with any appearance: `default`, `primary`, etc.

For now, I've added `opacity: 0.5` (along with `cursor: not-allowed`) as a visual cue for disabled state. We can revisit this after PoC


landitus:
LGTM!

PR #284, opened 2018-02-15
Add ButtonGroup component

siddharthkp:
1. Introduce a new component `ButtonGroup` which supports `compressed` prop
2. Reduced the padding for `ButtonWithIcon` (without text)

Results:

screen shot 2018-02-15 at 6 31 44 pm

Used it in `Form.Field` and `Form.Actions`

screen shot 2018-02-15 at 6 27 30 pm

screen shot 2018-02-15 at 6 27 51 pm

Using it client lists makes it slightly better than the current implementation, the one on production is also weird 😢. Should we use `compressed`  with `<Button link/>`?

screen shot 2018-02-15 at 6 25 48 pm

### DIFF
--- a/src/components/molecules/button-group/button-group.md
+++ b/src/components/molecules/button-group/button-group.md
@@ -33,7 +33,7 @@
 ```js
 <Stack>
   <ButtonGroup compressed>
-    <Button primary>Save changes</Button>
+    <Button>Save changes</Button>
     <Button>Clear</Button>
   </ButtonGroup>
   <ButtonGroup compressed>
--- a/src/components/molecules/button-group/button-group.md
+++ b/src/components/molecules/button-group/button-group.md
@@ -7,26 +7,39 @@
 
 ---
 
-## Examples
-
-```js
-<ButtonGroup>
+```jsx
+<ButtonGroup {props}>
   <Button primary>Save changes</Button>
   <Button>Clear</Button>
 </ButtonGroup>
 ```
 
+## Examples
+
 ```js
 <Stack>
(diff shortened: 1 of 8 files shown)


src/components/molecules/button-group/index.js
```diff
@@ -0,0 +1,46 @@
...
+  ${Button.Element}:not(:first-child):not(:last-child) {
+    border-radius: 0;
+  }
+`
+
+const StyledButtonGroup = styled.div`
+  display: inline-block;
+
+  ${Button.Element} {
+    margin-right: ${props => (props.compressed ? 0 : spacing.small)};
```
landitus: Let's go for spacing.xsmall here so they are a tad closer to each other

landitus:
> Using it client lists makes it slightly better than the current implementation, the one on production is also weird 😢. Should we use compressed with <Button link/>?

Once we `spacing.xsmall` tweak a few more I've got (up my sleave :) ) It will look even better than in production! I want to have these changes committed first so I can do the small tweak.

src/components/molecules/button-group/button-group.md
```diff
@@ -0,0 +1,45 @@
...
+    <Button icon="copy" />
+    <Button icon="trash" />
+  </ButtonGroup>
+</Stack>
+```
+
+```js
+<Stack>
+  <ButtonGroup compressed>
+    <Button primary>Save changes</Button>
```
landitus: Let's remove the primary button from this example, so they are both default. Usually when compressing the buttons, they have all the same importance.

Issue #65, opened 2017-12-19
Guiding principles of cosmos components to help make decisions


siddharthkp:
It's useful to keep all guiding principles in one place, these influence the smaller decisions we make 😄 

The aim of cosmos is to _help_ designers and developers deliver work faster and better.

&nbsp;

- _Common language for all teams._

	Code (components, tokens, etc.) and design assets (sketch files, icons, etc.) should always have the same naming convention/API and be kept in sync. The documentation is the tool to help bridge the gap between design, engineering and product. 

- _Encourage the best UX/UI practices._

	We should provide guidelines on when to use components and patterns (or not) to ensure consistency throughout our products (and within). 

- _Have meaningful defaults, bake convention in the components._

	Developers should be able to pick a component create the recommended design guideline out of the box. Optimise for ease of implementation. Bonus: Challenge conventions all the time.

- _Provide powerful abstractions while exposing the inner components at the same time._

	It should be easy + fast to build entire pages with cosmos, but at no point should the developer feel limited by the system. If there's a need to go out of the recommended way, they should be able to compose it with the inner/base components.

- _Be open about progress, direction, intentions._

landitus:
I think this should be in the homepage of the docs website

nkohari:
This is now captured in the docs, which is definitely a good spot for it! Any additional updates to it should go there.

PR #291, opened 2018-02-16
Added the guiding principles to the docs

landitus:
Following up on https://github.com/auth0/cosmos/issues/65, it's a good idea to put the guiding principles on the docs 😉

### DIFF
--- a/src/docs/docs-components/typography.js
+++ b/src/docs/docs-components/typography.js
@@ -11,6 +11,7 @@ const Heading1 = styled.h1`
 
 const Heading2 = styled.h2`
   font-size: 24px;
+  line-height: 1.4em;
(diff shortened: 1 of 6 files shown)


landitus:
While at it, re-ordered the sidebar to have the building blocks first and the compound components after. I think it makes more sense to have the simpler stuff first: Getting started --> Building Blocks --> Compound Components. 

wdyt?

siddharthkp:
> I think it makes more sense to have the simpler stuff first

I thought about it the other way round, folks should discover compound first because that's what we want them to use, they have patterns baked in. 

If a compound component does not do what you want, then you should fallback to building blocks.

landitus:
Oh! I like that approach too!

PR #298, opened 2018-02-16
TextInput: should use defaultValue instead of value

siddharthkp:
Saw a few warnings about using value and defaultValue together after https://github.com/auth0/cosmos/pull/205

We should use just defaultValue here, right?

@nkohari Ideas?


nkohari:
Yeah, you're right! Sorry about that. 😄

PR #307, opened 2018-02-16
Add example Create Client dialog to POC

### LINKED ISSUE #272: Create mockup of Create Modal Dialog for POC
Once the Page Header is merged (see #47, #201), we need to display a Create dialog with mocked-up content.

nkohari:
Closes #272.

This patch adds an example `CreateClientDialog` which is wired to the button in the `PageHeader`. It manages its state, but doesn't currently do anything with it on close. :)

screen shot 2018-02-16 at 3 02 43 pm

Obviously, it's missing the icons. Should we use the images currently being used in the Manage app, or were we planning to change them?


nkohari:
Forgot to mention, this patch also adds a `alignVertical` property to `<Stack>`, because I was originally trying to get the "cards" in the dialog to flex to the height of the largest. Unfortunately, I ended up punting on that and setting a fixed height (as is done in Manage now). I left the `alignVertical` property because I thought it might still be useful, but if we want to pull it out I can do so. :)

nkohari:
Added another commit that pulls the SVGs from Manage:

screen shot 2018-02-20 at 3 49 02 pm

The grays are slightly different but I wanted to use what was available in the tokens. :)

landitus:
I'll check it out tomorrow!

landitus:
Can you take a look @siddharthkp ? Would love to merge this :)

Issue #236, opened 2018-02-08
Improve display of header for Overview screen


nkohari:

nkohari:
Closed by #313.

PR #313, opened 2018-02-19
Added a new navigation and layout for the PoC Overview of Components

landitus:
This is a big one! But fortunately it doesn't affect components and it's contained to the Overview components. 

Now the Overview page will look like the screenshot. The idea is to be able to take a quick look at all components. But it could also be useful as a snapshot for each commit to be able to diff them until we have a better solution in place.

screenshot-localhost-3000-2018 02 19-16-27-45

Addresses this: https://github.com/auth0/cosmos/issues/236


nkohari:
Looks good to me! We may want to squash this into a single commit before merging to keep things simple. I'm happy to squash/rebase and fix the merge conflicts if everyone else is happy with this PR.

landitus:
I agree, I had a hard time rebasing this one yesterday, and now it’s outdated again :)

beneliflo:
Rebase done

landitus:
LGTM!

PR #316, opened 2018-02-20
List style

beneliflo:
Fix some styles on the list.

I add a prop for ButtonGroup called `align` as is used on Stack but this is useful for change the side of the margin.


src/manage/client-list.js
```diff
@@ -52,14 +77,13 @@ class ClientList extends React.Component {
               </StyledLogo>
```
landitus: My only comment would be to have the avatar and the name | type on the same stack group. This way, we can control the separation better.

landitus:
We could also add the avatar images for each client type, so it looks more real! wdyt @beneliflo ?

landitus:
@nkohari added those images in this PR today https://github.com/auth0/cosmos/pull/307 so we could re-use them here.

landitus:
Do you have an ETA on this one @beneliflo ? So we can have it before the demo :)

landitus:
I've made some small adjustments! 

- Added Link component for the main link and changed margin for avatar
- Added some notes as to-dos for a next iteration
- Updated lists examples in overview

LGTM :)

Issue #315, opened 2018-02-19
Better styles for Heading and Page Header


landitus:
- [x] Add line-height of 1.3 for `Heading` https://github.com/auth0/cosmos/pull/319
- [x] Add default margin for `Heading` to allow stacking with Paragraphs https://github.com/auth0/cosmos/pull/319
- [x] Add missing icon for PageHeader
- [x] Use `ButtonGroup` for the Page Header actions 
- [x] Fix styles for Client page header
- [x] Figure out how to override margin in PageHeader https://github.com/auth0/cosmos/pull/319
- [x] Figure out how to override margin in List (Connections Table) https://github.com/auth0/cosmos/pull/319
- [ ] Figure out how to add margin to `<Icon>`. Related to https://github.com/auth0/cosmos/issues/314
- [ ] Add documentation for PageHeader

nkohari:
Closing this, since it's a rollup of other issues. Some of those are fixed and the rest are captured elsewhere.

PR #319, opened 2018-02-20
Added default line-height and margin to Headings

landitus:
Started to address https://github.com/auth0/cosmos/issues/315
Still I’ve commented where we need to override the margin inside some components


beneliflo:
Added default line-height and margin to Headings AND margin to Paragraph :)

Issue #320, opened 2018-02-20
Masked state is using value and not defaultValue


landitus:
- [ ] Masked state is using value and not defaultValue
- [ ] Also, the default value when empty should be just empty :)

nkohari:
Closed by #328.

PR #328, opened 2018-02-27
Fixes for inputs

### LINKED ISSUE #308: Review: Fixed height for inputs
Review of changes in https://github.com/auth0/cosmos/pull/273, as some of them are not in master anymore

nkohari:
Closes #308 and #320.

Looks like the fixed height style got lost when the masked input mode was merged. This patch also makes the masked input mode use `defaultValue` for consistency's sake, but as discussed in Slack we may want to consider using `value` across the board if our goal is to work with controlled components.


landitus:
FYI: When reviewing this one I've noticed a few issues, but I maybe we could take care of them separately:

- https://github.com/auth0/cosmos/issues/330
- https://github.com/auth0/cosmos/issues/331

PR #355, opened 2018-03-05
Replace relative token imports with auth0-cosmos-tokens

siddharthkp:


siddharthkp:
Booting a container is timing out on now 😢, I don't think it's because of the changes in this file.

Issue #347, opened 2018-03-01
Best way to override the styles of components in the application level


landitus:
We did a bunch of **quick fixes** in this PR: https://github.com/auth0/cosmos/pull/343 that needed to override the styles of components inside the Manage PoC app. A margin here and there... and maybe more. I was wondering what’s the cleanest approach (with styled-components) applied so then we could move/integrate these overrides in future components updates. Or maybe, these styles at the app level will need to remain in the app.

Seems like something our users will run into as well… It was all inside the **Manage PoC** so, it’s all on the **application level**.

We used different approaches throughout:

### A. Wrapper with child selector

```
const Wrapper = styled.div`
  ${StyledElement} {
    position: absolute;
    right: 4px;
    top: 3px;
  }
`
```

It’s clean… But you need to wrap in a container element (wrapper) and also have the component export the styled version in  { }. It’s clean, but touches the original component code, which is not technically an override. People won’t be able to do it, unless we export the necessary styled component.

Looks like this technique suits our team’s case where we need to compose components into each other. I like it there.

### B. New styled component
In a case we needed to override the margin of Paragraph, I ended up with this…. which I didn’t like, but maybe it’s ok:

```
const StyledParagraph = styled(Paragraph)`
  margin-top: ${spacing.large};
  margin-bottom: 0;
`
 const Description = props => {
   return (

 <StyledParagraph>
       {props.children.text}{' '}
       {props.children.learnMore ? (
         <Link href={props.children.learnMore}>
           Learn more <ArrowMore />
         </Link>
       ) : null}

</StyledParagraph>
   )
 }
```

So you create another component, which has all the styles of Paragraph plus the overrides. I don't like to change the name of the component. It's difficult to name things... and to have even variants of the same thing seems difficult to track.

### C. Idea! [Made-up code]

Personally, I would like something more like this (this is all made up code!):

```
const overrides = `
  margin-top: ${spacing.large};
  margin-bottom: 0;
`
 const Description = props => {
   return (

 <Paragraph styles={overrides}>
       {props.children.text}{' '}
       {props.children.learnMore ? (
         <Link href={props.children.learnMore}>
           Learn more <ArrowMore />
         </Link>
       ) : null}

</Paragraph>
   )
 }
```

[...]

PR #369, opened 2018-03-06
Theme experiment

siddharthkp:
This is an experiment to see how much can we stretch cosmos to fit a different theme like extend.

Learnings:

1. `tokens` is the bomb

2. our tokens aren't the bomb yet, sometimes they refer to what they are and not what you should use them for. (had to invent generic `success` state to customise). We should attempt to make them more meaningful before we encourage others to use them.

3. some components like `Switch` have deeply nested components that might need customisation (hopefully rare use case, but super awkward: [#diff](https://github.com/auth0/cosmos/pull/369/files#diff-b2190fcb183cac4590a4f56e4e1ab53dR5). Adding more to https://github.com/auth0/cosmos/issues/347


nkohari:
@siddharthkp Are we planning to merge or continue to update this branch? If not, I'll close this PR for now.

siddharthkp:
I really want to, but maybe not the right time

PR #383, opened 2018-03-08
Allow form field description (helper text) to accept React components

### LINKED ISSUE #365: Form Field description prop should accept a React component
On Form Fields, the `description` prop (which shows the "helper text" beneath the field) must be a string. For example:

```jsx
const MyComponent = props => (
  <Form>
    <Form.TextInput label="Form label" description="test" />
  </Form>
);
```

Instead, we should expand this to also allow a React component to be passed. For example:

```jsx
const MyComponent = props => {
  const description = <Paragraph>Here is some <strong>rich-formatted</strong> helper text.</Paragraph>;
  return (
    <Form>
      <Form.TextInput label="Form label" description={description} />
    </Form>
  );
};
```

If the `description` value is a string, it should just be rendered as plain text.

nkohari:
Closes #365.

This was already unofficially supported, so this patch just removes the warning and updates the docs. There's still an unanswered question on the issue regarding whether we want to continue supporting backticks in strings, but this can be merged either way.

### DIFF
--- a/src/components/_helpers/reset.js
+++ b/src/components/_helpers/reset.js
@@ -1,4 +1,5 @@
 import { injectGlobal } from 'styled-components'
+import { fonts } from '../../tokens'
 
 const reset = () => injectGlobal`
   html, body, div, span, applet, object, iframe,
@@ -49,6 +50,10 @@ table {
 /* Our resets */
 * {
   box-sizing: border-box;
+}
+
+strong, em {
+	font-weight: ${fonts.weight.bold};
 }
 `
 
--- a/src/components/molecules/form/description.js
+++ b/src/components/molecules/form/description.js
@@ -2,7 +2,6 @@ import React from 'react'
 import styled from 'styled-components'
 
 import { spacing, colors } from 'auth0-cosmos-tokens'
-import Code from '../../atoms/code'
 
 const StyledDescription = styled.div`
   font-size: 13px;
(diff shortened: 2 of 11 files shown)


siddharthkp:
We should also delete the extra formatting inside https://github.com/auth0/cosmos/blob/master/src/components/molecules/form/description.js

landitus:
I think we could rename the prop to `helperText` as it describes the intent of this text better than description.

siddharthkp:
Bike-shedding now, but how about `instructions`?

landitus:
I think helpText is more standard, again, thanks to bootstrap: https://getbootstrap.com/docs/4.0/components/forms/#help-text <https://getbootstrap.com/docs/4.0/components/forms/#help-text>

siddharthkp:
> I think helpText is more standard, again, thanks to bootstrap: getbootstrap.com/docs/4.0/components/forms/#help-text <getbootstrap.com/docs/4.0/components/forms/#help-text>

They don't call the class/API help-text though:

```html
<label for="inputPassword5">Password</label>
<input type="password" id="inputPassword5" class="form-control" aria-describedby="passwordHelpBlock">
<small id="passwordHelpBlock" class="form-text text-muted">
  Your password must be 8-20 characters long, contain letters and numbers, and must not contain spaces, special characters, or emoji.
</small>
```

landitus:
I agree! It's just that we always refer to it as the "helper text", which is a common UI term that most designers will use for that text. I think bootstrap might be responsible for this, as many components got standardized by their wide adoption. In this specific case, I would call the prop `helpText` (or `instructions`) as they both convey the right usage.

nkohari:
Added a commit that removes the backtick replacement. (I wasn't sure if we still wanted that shortcut or not. I prefer it being removed, though!)

Did we want to rename the property, or can we merge this as-is?

landitus:
Looks better without the backticks for sure! Yeah, I’m pushing for renaming to helpText. I think it’s the best term for now

Issue #363, opened 2018-03-05
Remove Manage-specific functionality from PageHeader


nkohari:
The `<PageHeader>` component is a little too specific to the Manage PoC. We need to split it into a reusable component (which will be part of Cosmos) and a component that provides the Manage-specific functionality (which will be part of the example app).

landitus:
Solved in #385

PR #385, opened 2018-03-08
Org page header

beneliflo:


landitus:
Page header looks good! You should continue with the Clients page header (local component).

landitus:
Closes #363

Issue #380, opened 2018-03-07
Replace ClientType.Title with a Heading 4 in Create client dialog


landitus:

landitus:
Closes https://github.com/auth0/cosmos/pull/391

Issue #389, opened 2018-03-09
Margins for Headings and Paragraphs


landitus:
Headings and Paragraphs are one of the core components of a system. In most CSS frameworks (like Bootstrap), all typographic elements have margins (top and bottom) to allow for vertical rhythm while composing long form text. You can have titles, subtitles and paragraphs placed on the DOM and the correct separation would be achieved effortlessly. 

**Example of typographic elements:**
- Tutorials or Quickstarts
- An intro text after a Tab
- Sections and paragraphs between other components (Forms, tables)

Our initial approach had margin-less Headings and Paragraphs. But taking cues from these CSS frameworks techniques, we changed Headings and Paragraphs to have both top and bottom margin. 

The first issue comes from the first paragraph or Heading in a page, which doesn't need the top margin. So, we have overrides for margins at the page level. This is prone to error and maybe calls for a better solution. 

Note that we **do overrides** for these margins at the component level when needed, which is expected, so that seems ok.

I've been thinking about some strategies to avoid overriding the margins:

### Remove all margins, create a Section component 

Remove all margins of the typographic components and create a new parent component, called **Section** or **Text**, which will add proper margins to all Headings and paragraphs inside. This would allow to remove margin top to the first element (which-ever it is) and to the last. So that there's not an extra space below. 

```
<Section>
  <Heading size={1}>This is a title</Heading>
  <Paragraph>This is text</Paragraph>
</Section>
```

([Polaris does something similar](https://polaris.shopify.com/components/titles-and-text/text-container#navigation))

**Alternative: Keep only the margin-bottom, create a Section component**
Almost the same as A): Keep only the margin-bottom and add the margin top when used inside the Text component. The only change is that this is a little more consistent with our (current) spacing approach where all components only have margin-bottom to achieve spacing. 

--- 

I've researched other DS and have not found examples or docs of typography components. Just regular` <p>` and `<h1>` with plain CSS (with top and bottom margins).

Wdyt? Are there another options I'm missing out? Maybe there's a simpler solution! 🤔

landitus:
This idea should take care, among other things, of: 

```
// TODO: Need to figure out how to make the first Paragraph have no margin top. We could add a prop?
const FirstParagraph = styled(Paragraph)`
  margin-top: 0;
`

class Connections extends React.Component {
  render() {
    return (
      <div>
        <FirstParagraph>
          Connections are sources of users. They are categorized into Database,
          Social and Enterprise and can be shared among different applications.
[...]

Issue #384, opened 2018-03-08
Additional text styles (or components)


landitus:
We have a few more text styles (in addition to the Paragraph) that could be extracted as components that we can use inside other components:

- **Smaller paragraph**: Could be called **Caption** (as it is lighter and smaller)

<img width="672" alt="screen shot 2018-03-08 at 23 54 17" src="https://user-images.githubusercontent.com/239215/37188100-3ea52d10-232c-11e8-99ed-6d7ce6daa925.png">

<img width="154" alt="screen shot 2018-03-08 at 23 53 22" src="https://user-images.githubusercontent.com/239215/37188118-51c45e3e-232c-11e8-99d7-f85b45951686.png">

<img width="233" alt="screen shot 2018-03-08 at 23 53 33" src="https://user-images.githubusercontent.com/239215/37188111-4ab10912-232c-11e8-8058-63a28a4f7d18.png">

- **All Caps Text:** These are a very special style in Auth0 but it's used for some small titles and to denote the Type of entities in some parts of the application

<img width="501" alt="screen shot 2018-03-08 at 23 53 39" src="https://user-images.githubusercontent.com/239215/37188109-47459eb4-232c-11e8-8cdd-04f67abb7120.png">

<img width="326" alt="screen shot 2018-03-08 at 23 54 13" src="https://user-images.githubusercontent.com/239215/37188102-40f5e654-232c-11e8-9cad-ca6e14be11a1.png">

<img width="282" alt="screen shot 2018-03-08 at 23 53 18" src="https://user-images.githubusercontent.com/239215/37188126-5a32cd1c-232c-11e8-899e-5cceb12461ef.png">

With these new typography styles (plus Paragraph and Heading) we should be covered.

nkohari:
I guess I'm not sure what the difference between `Caption` and `Paragraph` is, from a visual standpoint. Also, I'd say `Caption` sounds like it's related to an image or diagram in some way.

For the small-caps styling on the text, maybe we need some sort of `<Text>` component after all which would allow a type, ex:

```jsx
<Text type="small-caps">
  ...
</Text>
```

I'm not sure each style merits having a separate component -- and if so, I have no idea what to call them! :)

Now that I'm thinking about it, a good litmus test for components might be whether we can come up with a name for them that effectively conveys their intended use.

landitus:
[...]

PR #391, opened 2018-03-09
update quickstart title to h4

beneliflo:
tracking https://github.com/auth0/cosmos/issues/380

### DIFF
--- a/src/manage/pages/clients/create-client-dialog.js
+++ b/src/manage/pages/clients/create-client-dialog.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types'
 import styled, { css } from 'styled-components'
 import { Dialog, Form, Paragraph, Stack } from 'auth0-cosmos'
 import { colors, fonts, misc, spacing } from 'auth0-cosmos/tokens'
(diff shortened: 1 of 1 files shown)


src/manage/pages/clients/create-client-dialog.js
```diff
@@ -67,13 +68,13 @@ ClientType.Image = styled.div`
   }
 `
 
-ClientType.Title = styled.div`
-  font-weight: ${fonts.weight.medium};
+ClientType.Title = styled(StyledHeading[4])`
```
nkohari: This is okay as a temporary hack, but if consumers of Cosmos have to alter styles like this, it's a pretty bad developer experience. 😞

nkohari:
I'm going to merge this, but as we continue to transform the PoC into an example app, I think we need to look at it very critically. Any "code smells" in the example app could be problems with the Cosmos API itself (the overriding of Heading styles is an example of that)!

landitus:
Agreed, this needs to be addressed with the new typographic styles: https://github.com/auth0/cosmos/issues/389 and https://github.com/auth0/cosmos/issues/384

PR #392, opened 2018-03-09
Created a Text component with 2 possible text styles

landitus:
Addresses: https://github.com/auth0/cosmos/issues/384


landitus:
Will do ASAP

nkohari:
I'm just going to merge this for now so the PR doesn't become stale, but we should add documentation eventually. :)

Issue #375, opened 2018-03-07
Extract Avatar component


landitus:
- [x] Extract from Manage PoC into Cosmos/components
- [x] Add size prop
- [x] Document
- [x] Consume it all across the PoC, clean up overrides

nkohari:
Closed by #398 and #401.

PR #398, opened 2018-03-12
add avatar component

beneliflo:
tracking https://github.com/auth0/cosmos/issues/375

### DIFF
--- a/src/components/atoms/avatar/avatar.md
+++ b/src/components/atoms/avatar/avatar.md
@@ -5,8 +5,6 @@
 
 `import Avatar from 'cosmos/avatar'`
 
----
-
 ```jsx
-<Avatar {props} />
+<Avatar image={<Logo />} {props} />
 ```
--- a/src/components/atoms/avatar/avatar.md
+++ b/src/components/atoms/avatar/avatar.md
@@ -1,0 +1,12 @@
+```meta
+  category: Actions
+  description: Avatar component contains images or symbols.
+```
+
+`import Avatar from 'cosmos/avatar'`
+
+---
+
+```jsx
+<Avatar {props} />
+```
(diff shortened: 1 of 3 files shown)


siddharthkp:
Some feedback for `Avatar`:

Need @landitus to chime in for design inputs

Bug
- If you supply a bigger image than `size`, it does not get _contained_

screen shot 2018-03-13 at 4 36 29 pm

API
- Instead of `image`, can we call it `source`? (Because it accepts a component not just an image)
- Also, what's the use of supporting component as source/image?

Design
- Should `backgroundColor` be customisable? I think we should lock that down for consistency.
- `size` as a number can quickly get out of hand, can we support 3 sizes: `[small, medium, large]` or maybe based on their purpose: `[icon, list, header]`. When we do the inventory exercise, we will can fine tune these values.

landitus:
- I like the idea of using small, medium, large for the `size` prop.
- Having `source` as a prop seems better.

for the background, I think we have 2 kinds of avatars:

- The `user avatar` (circled) The default has a letter and a random background color. If the user has an image, it will render the photo.

screen shot 2018-03-13 at 13 29 31

- The Avatar for an `entity` (which is this case): for an API or Client. It's a full image with no background. There are 4 default images for clients, depending on the type.   


screen shot 2018-03-13 at 13 31 07

Looking at it, the Avatar for entity should be **squared** to fit in APIs and other screens.

landitus:
I'm starting to think whether we need 2 different components for avatars, how they would be called or if they are just a single one 🤔

beneliflo:
Thanks for the feedback guys 😃 
Love it the idea of using only three size for avatars 👍 
`source` is the correct use, yep!

I don't think we should have 2 different components for avatars. The `user` avatar and the `entity` avatar both can be useful just using an avatar component.
That's where `backgroundColor` enters the game and we can use it in `user` avatars.

And I think we have to have this kind of discussion in an issue because PR's can be closed and deleted.

Issue #88, opened 2017-12-22
Binary props vs enum


siddharthkp:
What makes a more intuitive API?

```jsx
<Button primary large>Button</Button>

<Button transparent small>Button</Button>
<Button destructive small>Button</Button>
```

	vs.

```jsx
<Button type="primary" size="large">Button<Button>

<Button type="transparent" size="small">Button</Button>
<Button type="destructive" size="small">Button</Button>
```

The first approach has a simpler API, you write as you think, _I need a large primary button_, the conflicting props are not obvious, you need to rely on `prop-type` validation warnings to catch mistakes

![screen shot 2017-12-22 at 6 30 45 pm](https://user-images.githubusercontent.com/1863771/34312920-4c4677e8-e746-11e7-9fa1-5aea1e40da5d.png)

The second approach does not have conflict problems at the cost of a longer/more explicit API.

luisrudge:
I vote for the first one 😬

nicogarcia:
At first glance I like the first one but does it allow to dynamically set the value as does the second? E.g. `<Button type={buttonType} ...`

luisrudge:
it does.

```js
<Button
    transparent={shouldBeTransparent}
    small>Button</Button>
```

nicogarcia:
Yeah, but that should be `<Button transparent={shuldBeTransparent} primary={shouldBePrimary} secondary={shouldBeSecondary}...` I think, for some use cases. Maybe both approaches can be accepted? Or would it be confusing?

astanciu:
I too vote for the second one. You have one property (type), with many options (primary, transparent, etc..)

The first format would allow me to do bad things like:
```JSX
<Button primary transparent>Button</Button>
```

What's the type there?

luisrudge:
> Yeah, but that should be <Button transparent={shuldBeTransparent} primary={shouldBePrimary} secondary={shouldBeSecondary}... I think, for some use cases. Maybe both approaches can be accepted? Or would it be confusing?

Well, this is the code for your example:

```js
<Button type={shuldBeTransparent ? 'transparent' : shouldBePrimary ? 'primary' : shouldBeSecondary ? 'secondary' : 'default'} />
```

I rather have the first one 😊

alejofernandez:
I vote for option 1, I wasn't a fan of it when I saw it first, but I'm convinced now that it is more readable.

ramasilveyra:
I vote for `<Button primary large>` approach.

[...]

Issue #87, opened 2017-12-22
Should button be clickable while it's in loading state?


siddharthkp:
![screen shot 2017-12-22 at 6 10 12 pm](https://user-images.githubusercontent.com/1863771/34312499-62bbaad2-e743-11e7-831b-ec68b621163c.png)

If not, there are 2 ways of doing it:

1. Override `onClick` prop when the `loading` prop is passed
2. You need to pass `loading` and `disabled` both to disable clicks.

luisrudge:
As a future consumer, I expect this to be handled by the lib itself, so passing loading=true would automatically set to disabled (or ignore the onclick)

landitus:
I agree with @luisrudge !

nkohari:
This has been implemented in the current `Button` component.

PR #400, opened 2018-03-12
Replaced colliding Booleans with enums in Button and Tooltip

### LINKED ISSUE #364: Change use of Booleans to enums wherever they could collide
After some debate (see #88) we've decided to avoid the use of multiple Boolean props to control a single aspect of components, and instead use an enumeration. This will result in slightly more ugly code, but avoids potential issues where different properties could collide.

nkohari:
Closes #364.
Related to #88.

This patch combines the `default`, `primary`, `transparent`, `link`, and `destructive` props on `Button` into a single enumeration called `intent`. The name of this prop is debatable; there were several suggestions, so I just picked one that was more descriptive than `type`. 😄 

The `disabled`, `loading`, and `success` props are left as Booleans, and they influence (or override) the styles set by intents. It's still technically possible for these state-oriented properties to collide, but it's reasonable to have a combination of two. For example, you might want to disable a button while a form is saving, so you might set it both `disabled` and `loading`.

In `Tooltip`, the `top` and `bottom` props are combined into a `position` enum. We could just eliminate the `top` prop and keep the `bottom` prop as a Boolean. However, if we decided to add more positions in the future, it would either end up as a breaking change or result in an inconsistent API.

### DIFF
--- a/src/components/atoms/button/button.md
+++ b/src/components/atoms/button/button.md
@@ -30,10 +30,10 @@ Buttons are great to ask users for action
 <div>
   <Stack>
     <Button>default</Button>
-    <Button primary>primary</Button>
-    <Button transparent>secondary</Button>
-    <Button link>Clear</Button>
-    <Button destructive>Delete</Button>
+    <Button intent="primary">primary</Button>
+    <Button intent="transparent">secondary</Button>
+    <Button intent="link">Clear</Button>
+    <Button intent="destructive">Delete</Button>
   </Stack>
 </div>
 ```
@@ -59,7 +59,7 @@ Icon buttons work well in compact spaces. You can pick name of `icon` from [docs
 ```js
 <div>
   <Stack>
-    <Button link icon="copy" />
+    <Button intent="link" icon="copy" />
     <Button icon="copy" />
   </Stack>
 </div>
--- a/src/components/atoms/button/button.md
+++ b/src/components/atoms/button/button.md
@@ -30,10 +30,10 @@ Buttons are great to ask users for action
 <div>
   <Stack>
     <Button>default</Button>
-    <Button intent="primary">primary</Button>
-    <Button intent="transparent">secondary</Button>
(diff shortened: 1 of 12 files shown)


landitus:
Nice!!

I think we are going to have more positions for the tooltip, so it's nice to have the position enum!

landitus:
@nkohari Just a heads up, there are some docs and components in the overview that still have the old button syntax, like https://auth0-cosmos-enums.now.sh/docs/ButtonGroup

siddharthkp:
Looks really good!

Should the _state_ of the button be grouped together as well? 

Right now, one conflicting half is grouped, but the other conflicting half isn't.

`<Button intent="transparent" disabled loading success>Button</Button>`


Suggested: 
```jsx
<Button intent="transparent" state="disabled">Button</Button>
<Button intent="transparent" state="loading">Button</Button>
<Button intent="transparent" state="success">Button</Button>

```

siddharthkp:
>  but it's reasonable to have a combination of two. For example, you might want to disable a button while a form is saving, so you might set it both disabled and loading.

Combinations would definitely exist, but I don't consider that to be part of the API or job of the user 

Example: `<Button intent="primary" state="loading">Button</Button>` would be a blue button with the loading circle and disabled for clicks.

Related: https://github.com/auth0/cosmos/issues/87

nkohari:
Whoops! Sorry about that, I'll fix it momentarily.

nkohari:
While I was fixing the remaining problems, I realized that Form Actions have a property called `appearance` which we're just passing along to the `Button` as `intent`. Maybe `appearance` is the right name. 😄

siddharthkp:
I'd say pick one you like and we'll rename it later/whenever

beneliflo:
vote `appearance` 👍

landitus:
I like appearance too

nkohari:
Appearance it is!

Issue #376, opened 2018-03-07
Make the Client type options as radio buttons inside dialog


landitus:
We are using <a> tags for each option. Being a form, its more accessible to use a radio button group and style it. This is still a local component, shouldn't affect the library.

PR #403, opened 2018-03-13
[WIP] Change `Client Type` anchors to radio on dialog

beneliflo:
Tracking https://github.com/auth0/cosmos/issues/376

The principal idea what we talk with @landitus is change `anchors` to `radio` option in the selection of the `Client Type`. This is more effective in the use and the right semantic to this kind of options.
Maybe the engineering isn't my strong but is good to have to @siddharthkp 😄

This is an approach. Now this escalates and we have to build a component for this. 😅 

WDYT?

### DIFF
--- a/src/manage/pages/clients/create-client-dialog.js
+++ b/src/manage/pages/clients/create-client-dialog.js
@@ -8,17 +8,23 @@ import { StyledHeading } from 'auth0-cosmos/atoms/heading'
 import ClientTypeImages from '../../components/client-types-images'
 
 const ClientType = props => (
-  <ClientType.Element onClick={props.onClick} selected={props.selected}>
+  <ClientType.RadioElement onClick={props.onClick} selected={props.selected}>
+    <ClientType.Radio
+      type="radio"
+      value={props.name}
+      checked={props.checked}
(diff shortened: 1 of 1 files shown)


landitus:
Just want to add that these radio button disguised like cards are a potencial global component. We have them in use prominently in the onboarding. I'll post screenshots

siddharthkp:
Overall looks real good to me. 

With one minor code change:

I know this isn't your code (hi @nkohari!), but `Client.Element` or `Client.RadioElement` is a style that we use inside cosmos so that we can export it and override it.

I'd not want others to learn this pattern in the example app, it should be easy-peasy.

Because it's in the same file, I'd just replace that with 

```js
const RadioElement = styled.a``
```

and use that.

nkohari:
This has [WIP] in the title, so I'm assuming it shouldn't be merged? I added the `don't merge` label for the meantime.

nkohari:
@landitus @beneliflo Anyone know why this is still marked don't merge? Is this ready to go or should we abandon it? (See also #376.)

nkohari:
Closed since this has become stale. Will rework later using `Radio`.

PR #404, opened 2018-03-13
New avatar prop

beneliflo:
Tracking https://github.com/auth0/cosmos/pull/398 and https://github.com/auth0/cosmos/issues/406

Adding new props :D


siddharthkp:
Not sure if this the right place, but there's this bug: `If you supply a bigger image than size, it does not get contained`

nkohari:
Squashed commits and force-pushed before merge to keep history clean.

PR #420, opened 2018-03-20
Add size prop to Button

### LINKED ISSUE #417: Add support for large-sized Buttons
Currently in Manage, there is a large button in addition to the regular-sized ones that we already support. In order for Cosmos to be used in Manage, we need to add a `large` Boolean prop to `Button`.

@landitus could you provide mockups/screenshots of the "large" `Button`?

nkohari:
Closes #417.

This patch adds a `size` prop to `Button`, which controls the vertical height and horizontal padding of the button. Also, the branching logic in `Button` was quickly turning to spaghetti so I took a stab at simplifying it. In addition to (hopefully) making the code easier to understand, this should also fix some bugs with edge cases; for example, previously setting a button with only an icon to `success` would cause it to gain a minimum width and horizontal padding.

I also added the new large button to `PageHeader` and `EmptyState`. I think we probably want the `small` button inside of the clients table in the example app, but I wasn't sure.

### DIFF
--- a/src/components/atoms/button/button.md
+++ b/src/components/atoms/button/button.md
@@ -13,18 +13,20 @@
 
 ### Basic button
 
-Buttons are great to ask users for action
-
 ```js
 <Button>Save</Button>
 ```
 
-### Button types
+### Button appearance
 
-* Primary actions help attract attention to the main action
-* Use secondary actions for actions that are not as important
-* Link button for linky stuff, idk?
-* Destructive buttons warn the user about it's effects. Proceed with caution.
+The `appearance` prop defines the overall visual style of the Button. You can use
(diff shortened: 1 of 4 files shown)


landitus:
> I think we probably want the small button inside of the clients table in the example app, but I wasn't sure.

Not sure either, will check

nkohari:
Added a commit to change the spacing on the large buttons per conversation w/ @landitus.

landitus:
LGTM! I think we'll need to update the Overview with more sensible examples after this. But I could do it in another PR.

landitus:
Fixed one of the examples in the docs!, LGTM!

PR #422, opened 2018-03-22
Restructure part 0: Transpile components

siddharthkp:
This is an ugly PR, but needed for restructuring.

Things will not be so bad at the end.


siddharthkp:
Superseeded by https://github.com/auth0/cosmos/pull/430

PR #425, opened 2018-03-23
Add new "link mode" and new "cta" apperance to Button

### LINKED ISSUE #423: Replace link appearance in Button with "link button" mode
Currently, if `link` is specified as the value for the `appearance` prop on `Button`, we are rendering an HTML `<button>` whose style resembles a hyperlink. In order to support the use case of navigating the browser when a user presses a button, we want to replace this `link` appearance with an actual "link button" mode.

If a user specifies an `href` property on a `Button`, instead of rendering a `<button>` element, the `Button` should render an `<a>` with the appropriate `href`. This mode should also support the a `target` property with the same options as in the `Link` component, the value of which is passed through to the `<a>` element.

### LINKED ISSUE #424: Add CTA (call-to-action) Button appearance
Previously called "primary alt", the main buttons used in page headers are a branded orange color. We want to support these buttons in Cosmos as a new appearance called `cta`.

nkohari:
Closes #423.
Closes #424.

This patch adds a new "link mode" for `Button`, which will cause it render as an `<a>` instead of a `<button>` if an `href` is specified. The larges change here is switching from `height` to `line-height` to control the vertical sizing of the button. This was necessary to allow the sizing to be the same whether the `Button` is rendered as an `<a>` or as a `<button>`.

This patch also adds a `cta` appearance to `Button`, and renames `transparent` to `secondary`. We were originally planning to remove the `link` appearance entirely in favor of the "link mode"; however, we decided to keep it since we are using `link` appearance inside of `FormField` actions.


landitus:
Can't see the staging, can you?

landitus:
Got to check it out on my local machine. LGTM!

PR #429, opened 2018-03-26
Add dummy app for size testing

siddharthkp:
| file | size |
| ------------- |-------------| 
| just-react.js | 32.0 KB |
| with-styled.js | 46.1 KB |
| with-cosmos.js | 64.4 KB |
| with-part-cosmos.js | 47.4 KB |

- The cost of an app that renders `<h1>Hello world</h1>` is 32 KB
- add styled components `styled.h1` and it jumps to 46.1 KB
- If you add `{ Heading } from cosmos`, it adds ~ 18 KB!
- If you directly import `Heading from cosmos/atoms/heading` it adds just 1.3 KB (part-cosmos)

Without tree shaking the weight is very visible, a lot of that weight is coming from `icons.json`: 7KB

The total cost of cosmos right now is ~ 32KB which is definitely on the higher side.

- ~14KB is coming from styled-components and it's dependencies. We can invest some time on it and help them minimising the size.

- We shouldn't obsess over the size of each component because cosmos would be adopting as a whole (most of the time)

- For custom webpack configs, promoting tree shaking is the best solution

- For apps with create-react-app, we should try to minimise the cost of icons 🤔 Will circle back to this later.


siddharthkp:
webpack 4 makes it super easy to treeshake 😄 

| file | size |
| ------------- |-------------| 
| just-react.js | 32.0 KB |
| with-styled.js | 45.9 KB |
| with-cosmos.js | 47.1 KB |
| with-part-cosmos.js | 47.1 KB |

PR #430, opened 2018-03-26
Add publish script

siddharthkp:
Problem:

1. tokens, babel-preset and components should be versioned + published together

    We don't need them at different versions. With a team our size, it's easier to maintain a tightly bound system.

1. I've really grown to like the yarn workspaces

    It let's us write `import { colors }` from 'auth0-cosmos/tokens' instead of `../../../tokens`. Easier to write, easier to _grep_. I'd like to keep that.

&nbsp;
> Good developer experience leads to good user experience.

&nbsp;

Solution: 

Sometimes the worst solution is the best solution 😋 
I wrote a script that organises the repo to a more publish friendly setup 😄 

- create `dist` folder
- copy `tokens`, `babel-preset` and `components` _packages_ to it
- transpile components just enough
- copy version from repository root `package.json` to all _packages_
- publish them one by one.

&nbsp;

The whole thing runs by typing `yarn run publish` (instead of `yarn publish`)

&nbsp;
TODO: 
✅ `icons.json` is not handled yet.


siddharthkp:
@nkohari Why do we gitignore `src/components/atoms/icon/icons.json`, I assume this isn't a file that changes often, can we add it to version control?

siddharthkp:
> yarn deploy to avoid any confusion with the built-in yarn publish command.

Considering release branch is also our deployment branch. It makes sense!

> That said, it might be good to add it to the pre-commit hook to ensure it can't get out of sync. I think that was the reason it was ignored in the first place.

love it!

> Alternatively, we could add the build script as a post-install hook, so icons.json is generated when the consumer installs the package, but I don't know if that's necessary or better in any way. 😄

That's a good idea. There is a lot of magic that happens on `npm run dev`, we should setup most of it in `postinstall` as well.

PR #435, opened 2018-03-26
First pass at a changelog

### LINKED ISSUE #361: Create changelog
We need to publish a changelog to inform users of changes that we make to Cosmos, especially to the API. The changelog should be part of the documentation site.

For Cosmos 0.1, we need to create the first version of this changelog, and decide on how we want to maintain it going forward.

nkohari:
Closes #361.

This patch adds a `changelog.md` file to the root of the repo, which should be kept up-to-date with important PRs that are merged. If you know you're introducing a significant addition or change (especially breaking changes) it's probably a good idea to update the changelog in the commit itself. Before we ship new versions we should always do a final pass to ensure the changelog contains all the things we want to highlight.

The contents of the changelog is also added to the docs at `/docs/changes`. Right now, I added a (rather hacky) step that reads the contents of `changelog.md` and writes it to `src/docs/markdown.json` so it can later be read by `markdown-to-jsx` -- basically the same thing that the component documentation does now.

I started getting a little too fancy with this, trying to parse the Markdown to improve the styling of the changelog in the docs, but I backed off once I realized how brittle it was going to be. Unfortunately, that means that right now we're also sort of limited in how we can display the content in the docs. I'd like to do things like create anchors on each version, etc. but right now it's not easily possible.

I think there could be some merit to "inverting" the process, and hand-authoring a JSON file of commits and generating the Markdown for `changelog.md`... but I wasn't sure if that would become cumbersome to edit. 😄

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,0 +1,19 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## 0.1.0 [March 27, 2016]
+
+### Added
+
+* New `cta` appearance for `Button`, to be used for primary calls to action. [#425]
+* New "link mode" for `Button`. Passing an `href` prop to `Button` will cause it to render as an `<a>` instead of a `<button>`. [#425]
+* _Unstable:_ New `Avatar` and `Thumbnail` components. [#398]
+* _Unstable:_ New `Text` component for styled text. [#392]
+
+### Changed
+
+* **BREAKING:** Colliding Boolean props governing appearance of `Button` were replaced with an `appearance` enum. [#400]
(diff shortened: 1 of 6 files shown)


landitus:
Nice, I see your points. Love the idea of anchors as I wanted to have them for the docs as well. I think we'll be able to improve the presentation with some tags and other fancy things :) but we'll still need to put react and mdown together (right?). This is something we can tackle as a separate project to improve the docs and how we publish content there.

siddharthkp:
Question (not critique): What's the benefit of using a file over using [github releases](https://github.com/auth0/cosmos/releases)?

nkohari:
Since we are planning to publish to npm, I didn't think we would also want to publish releases to GitHub. I guess I didn't even realize that you could publish a release to GitHub without attaching build artifacts. 😄 

My thinking was that the main place we'll want to share the changelog is in the docs, but that it would also be a good idea to add a regular file to the repo for people who are browsing the source. If we're interested in publishing releases to GitHub, i could change the piece in the docs that reads the Markdown file to read from the GitHub API instead...

siddharthkp:
I'm honestly not sure. I guess I'm used to go to the releases tab on GitHub for open source stuff, but because we have a closed audience, we can make the rules :)

nkohari:
We could do all three, to be honest. If we want to make GitHub releases the source of truth, we can build both the Markdown file and the docs section from it by calling the API.

Issue #427, opened 2018-03-23
Review button states colors


landitus:
<img width="972" alt="screen shot 2018-03-26 at 12 50 30" src="https://user-images.githubusercontent.com/239215/37923534-b913b45e-3105-11e8-8d93-729d1e958342.png">

PR #436, opened 2018-03-26
Buttons styles (and tokens)

beneliflo:
Closes https://github.com/auth0/cosmos/issues/427


landitus:
Found one small bug: the disabled buttons still get hovered (and maybe focus). As a disabled button, it shouldn't react to user actions. Should we use a `pointer-events:none;` ? Also could be used when the button is in 'loading' and 'success state'

beneliflo:
solved

Issue #377, opened 2018-03-07
Extract ClientsList from clients.js


landitus:
Let's extract the Clients List as a local component so that it's encapsulated. This way, the PoC is more polished and shows a good implementation and practices.

nkohari:
Closed by #441.

PR #441, opened 2018-03-27
Improve client list

beneliflo:
Tracking https://github.com/auth0/cosmos/issues/377

Separate client list into a component.


src/manage/pages/clients/index.js
```diff
@@ -108,25 +41,12 @@ class ClientList extends React.Component {
         />
         <List>
```
landitus: Let's keep the entry point of this List to one single local component/file. For example <ClientsList /> This way, we keep this page simple and componentized. Also, this could help us show how we think local components could be implemented in our PoC.

PR #442, opened 2018-03-27
Improve connection list

beneliflo:
Separate connection list into a component


src/overview/examples/lists.js
```diff
@@ -77,38 +78,12 @@ const Type = styled.div`
...
-  span {
-    display: inline-block;
-    vertical-align: middle;
-  }
-`
-
 const ListsExample = () => (
   <Section>
     <Example title="List of Connections">
       <List label="Social">
```
landitus: Let's keep the entry point of this List to one single local component/file. For example `<ConnectionsList />` This way, we keep this page simple and componentized. Also, this could help us show how we think local components could be implemented in our PoC.

PR #449, opened 2018-03-28
Enable push events for travis

siddharthkp:
Fixes https://github.com/auth0/cosmos/issues/446

chromatic needs push events
now-cd needs pull_request events

¯\_(ツ)_/¯

I'm enabling both and making sure they don't fight 😛

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,3 @@ after_success:
   - cd build && npx now-cd --team=auth0-design
 notifications:
   email: false
-branches:
-  only:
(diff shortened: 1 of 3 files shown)


siddharthkp:
I knew this would happen so `now-cd` already handles it:

```
$ cd build && npx now-cd --team=auth0-design
 WARN  now-cd only works on push events
```

But we get duplicated chromatic builds 👎

siddharthkp:
`if-env` 💪

PR #455, opened 2018-04-02
Apply line-height only to sc- components

siddharthkp:
The only difference between the resets of the styleguide and cosmos is `line-height`

We want cosmos components to have our line-height, but not break everything else, hence as a hack, we're setting it on styled-components elements.
 
Note: This will break on applications that already use styled-components


siddharthkp:
@landitus Need a review of the components to make sure nothing unintentional happened. chromatic thinks so.

landitus:
Let me check it out

landitus:
I need to fix some numbers to make the buttons look ok with the new line height and also fix a design bug I caught today. I need to make sure the buttons and inputs align properly when displayed in the same line

landitus:
Got the changes but I think they exceed the scope of this PR. @siddharthkp lmk if you still want me to push the changes to this branch.

siddharthkp:
@landitus Updated the tests, now there are 6 files

landitus:
I have some conflicts when merging these changes, so I'll need help @siddharthkp

landitus:
I've solved the conflicts, so it's up for review!

PR #484, opened 2018-04-09
sketch library: basic setup

siddharthkp:
This is a brittle setup that uses 2 other brittle tools 😅

Pipeline:

1. We write [react-storybook](https://github.com/storybooks/storybook) stories where each symbol is a different story. [See example](https://github.com/auth0/cosmos/pull/484/files#diff-0a8afbb556d886d4e52edb09a11cefaf)
2. Run `yarn sketch`
	2.1 It builds a static storybook
	2.2 This storybook is imported by [story2sketch](https://github.com/chrisvxd/story2sketch) which uses  + [html-sketchapp](https://github.com/brainly/html-sketchapp) to create a `asketch`file
	2.3 We pass this file to `tooling/cleanup-sketch.js` to manipulate the json to our liking
3. This file can be loaded into sketch using the [Almost-Sketch-to-Sketch](https://github.com/brainly/html-sketchapp/releases/tag/v2.0.2) plugin
4. The file ins't perfect, we need to replace the fonts of each symbol manually 🙄to Fakt Pro
5. Save this file as `cosmos.sketch`

We have a library now!

Some of these things can hopefully be pushed into the cleanup-script


kdzwinel:
Just saw this! Excited to see you guys use html-sketchapp ♥️ 

You should be able to avoid replacing fonts manually and move this to your cleanup script, we (Brainly) do this: https://github.com/brainly/html-sketchapp-style-guide/blob/06722b663e15eaeef32491ce388e466f10ae43ae/src/styleguide2asketch.js#L56-L58 (although we don't use story2sketch). The root of the issue is that names of the fonts in your css don't match names used by your OS.

chrisvxd:
+1, great to see you using story2sketch.

I have plan to add better viewport support to story2sketch, which would allow you to remove the logic in the cleanup script once it lands.

Tracking here: https://github.com/chrisvxd/story2sketch/issues/3

chrisvxd:
Custom viewports have landed in `story2sketch@1.0.0`, so please upgrade in your own time.

Issue #259, opened 2018-02-12
Expressive example in prop table


siddharthkp:
The format for prop type is like this.

```jsx
<Tooltip {props}>Hover over me</Tooltip>
```

&nbsp;

In the case of required props without defaults, the example looks incomplete. Example: `content` in `Tooltip`

```jsx
<Tooltip {props}>Hover over me</Tooltip>
```

results in this hanging black tooltip:

<img width="104" alt="screen shot 2018-02-12 at 10 05 51 am" src="https://user-images.githubusercontent.com/1863771/36083753-562cf642-0fdc-11e8-8db4-72ed418683f8.png">

&nbsp;

The workaround right now is hardcoding the prop (kinda):

```jsx
<Tooltip content="This is a placeholder" {props}>Hover over me</Tooltip>
```

The problem with this approach is you end up with duplicated props when you use the prop table. The latter prop is the one that's picked up, (magic!) but it does look ugly.

<img width="790" alt="screen shot 2018-02-12 at 10 08 40 am" src="https://user-images.githubusercontent.com/1863771/36083785-bbc49e7e-0fdc-11e8-9694-92ff44a41f94.png">

Proposal:

Being able to add default values for the example as a prop (object) which only works in the docs and does not reach the document.

```jsx
<Tooltip  {props} defaults={ { content:"This is a placeholder", top: true } }>Hover over me</Tooltip>
```

This is also magic but if it is predictable for us (documentation writers), we can hide the ugly parts from users (documentation readers)

siddharthkp:
@landitus @beneliflo Thoughts?

landitus:
> Being able to add default values for the example as a prop (object) which only works in the docs and does not reach the document.

- What do you mean by `document`?
- `defaults={ { content:"This is a placeholder", top: true } }>` won't be shown in the code editor part, right?

siddharthkp:
> What do you mean by document?

I think I meant those are the defaults for documentation only, they are not part of `Component.defaultProps`

> defaults={ { content:"This is a placeholder", top: true } }> won't be shown in the code editor part, right?

Nope, we gobble that up

landitus:
mmm, I see. We'll need to find another solution then.

landitus:
So, maybe this could work!! and solves also https://github.com/auth0/cosmos/issues/488

PR #491, opened 2018-04-10
Add {props} to emptyState example

### LINKED ISSUE #488: Empty State propTable won't update on helpUrl prop
screen shot 2018-04-09 at 20 51 33

Checking the Empty State propTable, I've noticed it won't when addind the helpUrl prop. It seems like an issue with the propsTable component.

siddharthkp:
Fixes #488

The docs are more interactive now, but not copy-friendly because of the repeated props (see staging)

This is a bigger problem, we don't have a way of adding defaults props for just the docs, details: https://github.com/auth0/cosmos/issues/259

I'm not sure if this change is n+1 or n-1, what do you think?


landitus:
I don't see the repeating props. Looks fine to me!

screen shot 2018-04-10 at 21 09 33


We can also discuss #259. Sorry I totally missed it!

siddharthkp:
> I don't see the repeating props

Try adding icon or title

nkohari:
Yeah, I think your instincts were right on this @siddharthkp ... I'm not sure this is a good change unless we can find a way to eliminate the redundant properties. It definitely looks like a bug in the docs, and I think it will be confusing if someone copies and pastes it. (We'll definitely get bug reports about it. :D)

nkohari:
I think we have consensus that this shouldn't be merged as-is, and it's a larger issue that might need some refactoring of `Playground`. Unless anyone has a problem with it, I'm going to close this PR and we can circle back to it another time.

siddharthkp:
@nkohari Don't know if it's too late, but https://github.com/auth0/cosmos/pull/496 should make this PR possible?

siddharthkp:
Now, this PR is n + 1

Issue #479, opened 2018-04-06
Proposal: Tokens v2


siddharthkp:
Prior art: https://github.com/auth0/cosmos/pull/176

We want tokens to do 2 things for us:

1. Hold design decisions
2. Provide an alternate adoption path for non-react apps

We can achieve both by creating the right abstraction and naming things nicely. Basically, we're screwed.

I'm kidding, we got this. I'll take the example of color because it's easy to talk about.

### Layer 1: Base

This is the hard coded hex color information

```js
const colors = {
  blue: {
    lighter: '#14AAFF',
    light: '#4e9ee2',
    base: '#2386DB',
    dark: '#1f78c5',
    darker: '#1c6baf'
  }
}
```

### Layer 2: Purpose/Decisions

This is where our design decisions live, this is what makes a product _on brand_. We'll also need to document these decisions with reasons, DOs and DON'Ts.

```js
const conventions = {
  textOnDark: colors.white.base,
  textOnLight: colors.gray.darkest,
  primaryAction: colors.blue.base,
  destructiveAction: colors.red.light
}
```

### Layer 3: Components/Usage

Components follow the conventions described above. How components use the convention doesn't need to live in conventions, it can live in components. This makes them more flexible.

```js
const PrimaryButton = styled.button`
  background: ${ conventions.primaryAction };
  color: ${ conventions.textOnDark };
`

const Link = styled.a`
  color: ${ conventions.primaryAction };
`

const DestructiveAction = styled.button`
  background: ${ conventions.destructiveAction };
  color: ${ conventions.textOnDark }; 
`

const TrashIcon = styled(Icon)`
  background: ${ conventions.defaultAction };
  path {
    fill: ${ conventions.destructiveAction };
  }
`
```

landitus:
I like the 3 levels! My only concern is with the naming strategy. Like `primaryAction` doesn't give any context. Maybe `background.primaryAction` makes more sense. This will allow us to place variables correctly, and not use a bg color for text inside our components. 

But the layering notion is good and we should move forward with it, so we can work on the **conventions naming and structure**

Other than that, we'll need to think about:
- Exporting to other formats
- Using helpers to manipulate color and numbers

siddharthkp:
Moved to discussion: https://github.com/orgs/auth0/teams/design-systems/discussions/8

PR #498, opened 2018-04-12
Tokens v2 example (feedback needed)

siddharthkp:
1. Read the proposal and comments here: https://github.com/auth0/cosmos/issues/479

2. Already found colors that were in tokens and not used, the new structure prevents that

3. Chromatic tests passed = no breaking changes

4. This is still a work in progress, the code is nowhere close to what it should be like

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "src/components",
     "src/docs",
     "src/tokens",
+    "src/tokens-v2",
     "src/codemods",
     "src/overview",
     "src/manage"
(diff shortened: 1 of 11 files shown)


landitus:
I'm going to use a bunch of these tokens inside within this branch to test the new setup. Once we have a good idea if they work, we'll merge to master.

siddharthkp:
@landitus How was your experience using these?

siddharthkp:
How can I make this pull request easier to review 😄

nkohari:
If this is in a usable state, I think we should just merge it. This is the second PR for the second level of tokens, and I'd hate to abandon it again. :)

siddharthkp:
This PR is only good for example, it doesn't implement enough tokens/components

We can close this

landitus:
Sorry for the delay. Like it a lot! Let's move this forward!

Issue #394, opened 2018-03-09
Proposal: List Component


landitus:
We've been looking with @beneliflo for a better approach to the List. Inspired by the API and Dev experience of the Forms, we think that closing and controlling the layout will benefit everyone:

```jsx
<List
  widths={(30, 30, 40)}
  renderItem={item => (
    <List.Item>
      <List.Item.Meta
        avatar={ <Avatar src="avatar.png" /> }
        title={<a href="https://www.google.com">{item.name}</a>}
        type="Mobile Application"
      />
      <List.Item.ID content="Dsq0xjhJZAD7RvezvfjCDUip"></List.Item.ID>
      <List.Item.Actions actions={[
		{ label: 'Delete', icon: 'delete', method: () => {} },
		{ label: 'Edit', icon: 'edit', method: () => {} },
		{ label: 'Copy', icon: 'copy', method: () => {} },
	  ]} />
    </List.Item>
  )}
/>
```

This would require us to build all the possible internal components to render: switches, avatars, button groups, etc.

The idea would be to be able to achieve most of these designs (but maybe some lists will become data tables (for pagination and sorting) o we'll more specific components, like the key-value editor which has inline editing.

![lists and tables](https://user-images.githubusercontent.com/239215/37231863-803e2980-23cb-11e8-8426-3b25664d6adf.png)

----

This is the original approach:

```jsx
<List>
	{clients.map(client => (
	  <Stack key={client.id} widths={[35, 40, 25]}>
	    <NameGroup>
	      <StyledLogo>
	        <Avatar image={client.image} />
	      </StyledLogo>
	      <div>
	        <Link href={`/clients/${client.id}`}>{client.name}</Link>
	        <Type>{client.type}</Type>
	      </div>
	    </NameGroup>
	    <Stack align="left">
	      <ClientID>Client ID</ClientID>
	      <Code>{client.id}</Code>
	    </Stack>
	    <ButtonGroup align="right">
	      <Button icon="quickstarts" label="Quickstart" />
	      <Button icon="settings" label="Settings" />
	      <Button icon="code" label="Addons" />
	      <Button icon="connections" label="Connections" />
	    </ButtonGroup>
	  </Stack>
	))}
</List>```

nkohari:
I think I understand what you're going for, but the API is sort of confusing to me. Can you help me understand what the `List.Item.ID` and `List.Item.Meta` sub-components are for?

[...]

PR #504, opened 2018-04-13
Add ResourceList and ResourceListItem components

nkohari:
This patch implements the beginning of the `ResourceList` component discussed in #394.

The documentation still needs to be improved, and I haven't replaced the existing implementation in Client List, but I wanted to share it to get feedback.

### DIFF
--- a/src/components/atoms/link/index.js
+++ b/src/components/atoms/link/index.js
@@ -28,3 +28,4 @@ Link.defaultProps = {
 }
 
 export default Link
+export { StyledLink }
--- a/src/components/atoms/thumbnail/index.js
+++ b/src/components/atoms/thumbnail/index.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import styled from 'styled-components'
 import PropTypes from 'prop-types'
-
+import { Image } from '@auth0/cosmos'
 import { misc, colors } from '@auth0/cosmos-tokens'
 
 const sizeContent = {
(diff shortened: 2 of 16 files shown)


landitus:
Looks good! I just caught a visual regression with the List. It’s about the action buttons position and the width of columns. Could we add the test to chromatic before moving forward? So we test it against the current design on 0.0.1

nkohari:
Yep, that's a good idea! Like we were saying, good for us to get in the habit of making sure we have coverage before merging. 😀

landitus:
So far I've seen one visual bug, which might be related to https://github.com/auth0/cosmos/issues/517

The tooltip for the button group is not centered to the button. It looks like it's using the spacing of the margin as well.

nkohari:
Hmm... you're right. It isn't a regression in this patch though; here's the behavior with the old `ClientList` on `master`:

screen shot 2018-04-24 at 8 37 53 am

I'll create a new issue to fix this in another patch, but it shouldn't stop us from merging this.

landitus:
Sounds good.

Sent from Astro <https://www.helloastro.com> for Mac

On 24 Apr 2018 at 09:39, Nate Kohari <notifications@github.com> wrote:


Hmm... you're right. It isn't a regression in this patch though; here's the
behavior with the old ClientList on master:

[image: screen shot 2018-04-24 at 8 37 53 am]
<https://user-images.githubusercontent.com/1576/39187274-d989c268-479a-11e8-9b62-099770b35554.png>

I'll create a new issue to fix this in another patch, but it shouldn't stop
us from merging this.

—
You are receiving this because your review was requested.
Reply to this email directly, view it on GitHub
<https://github.com/auth0/cosmos/pull/504#issuecomment-383915001>, or mute
the thread

Issue #489, opened 2018-04-10
Add Radio buttons component 


landitus:
Design reference:

<img width="1030" alt="screen shot 2018-04-18 at 13 54 30" src="https://user-images.githubusercontent.com/239215/38946693-a4556da6-4310-11e8-942c-fd6de6ae1196.png">

PR #511, opened 2018-04-17
Radio component

beneliflo:
Fixes https://github.com/auth0/cosmos/issues/489 

Trying to build Radio component and RadioGroup :)


landitus:
Good! A few comments:
- I think we could wrap the Radio buttons in a `RadioGroup` tag. This way we'll be able to add props for layout and also control the separation between items.
- Also, can you add the Radio Button to the form? This way we can check how it aligns with the labels and the multiple layouts of form.
- @beneliflo I'm updating the issue with a better screenshot of the radios design and I've created a sketch file so you can inspect the design.

nkohari:
Some feedback:

1. Putting `Radio` in atoms and `RadioGroup` in molecules makes the docs a little confusing. It took me a minute to realize there was a `RadioGroup` at all. 😄 Unless there's a reason a user might want to use a `Radio` by itself, I think we should add it as a "sub-component" to `RadioGroup`, similar to `Form` -- although the component should just be called `Radio`, not `RadioGroup.Radio`.
2. I see what you were saying about `defaultChecked`. I think we may want to have a broader conversation about whether we want our components to be controlled (stateless) or uncontrolled (stateful). Trying to support both is going to be a nightmare, and I think we'll have to support stateless components anyway. If the `Radio` was stateless, all you'd need to do is pass the `checked` prop when the value of the `RadioGroup` matches the value of the `Radio`.

Otherwise, looks good!

landitus:
I like the idea of having a Radio subcomponent (and not `RadioGroup.Radio`). When you want to use them, you just import one component and you get both.

Regarding being stateless, agree it should be our default way, at least until we decide otherwise.

nkohari:
I implemented the changes I suggested in my previous comment; now `RadioGroup` is attached to `Radio`, so it can be used via:

```jsx
<Radio.Group>
  <Radio ... />
  <Radio ... />
  ...
</RadioGroup>
```

I've also made the components stateless, and they accept an `onChange` callback to allow a container component to manage the state externally. Also, I changed `isChecked` to `checked` to match the name of the property in HTML... I figured that would be a source of confusion otherwise. 😄

landitus:
One thing I've just noticed is that by having a single component in the docs, we lose the props and usage of the `<RadioGroup>`, which is very handy as it provides layout and other properties for controlling the Radio children. 

Maybe we can sort this out in the documentation, by having Radio Group be a subitem in the navigation, so we can keep the page and the props table. It sounds awkward, but it's an improvement.

Also, noted that by switching to stateless, the radios in the docs are no longer interactive (can't check them). This at first looks like a bug 🐞, so I was thinking:
- Is there a way we can showcase them properly in the docs (for all examples)? We could try adding boilerplate code in the examples.
- Let's document that they are stateless and maybe from the above question, we can provide in the docs an example of how to use them properly.

I think this is all related to how we document the component, more than the code itself. 
wdyt?

landitus:
Looks good! 
- I think it would be best to add the state behavior (the handleChange function) in all the Docs and also in the examples in the Overview. Even in the Props table representation would make sense. It's just that without the state, it feels like the component doesn't work properly.
- We could add a radio to this form here: https://auth0-cosmos-radio-component.now.sh/docs/form-field#examples

landitus:
> Either way, from the perspective of the docs, I think we should always demonstrate controlled components, because otherwise it isn't always clear how to interact with the component.

Yep, my comment referred to the docs part. Just makes sense to showcase it with a controlled component. Then the Radio component you import from Cosmos remains as it's now.

nkohari:
That makes sense. I'm saying I think we should look at all of our docs and make sure that we're demonstrating controlled usage, if that's the main use case for the components.

landitus:
Gotcha. When using the Radio in a form (with out state) it looks weird, so it's something to look at too.

nkohari:
Since that conversation is probably a broader one, I'm going to merge this as-is.

landitus:
Ok!

PR #513, opened 2018-04-19
[Hotfix for 0.1.2] Ignore null children in Tabs instead of throwing error

### LINKED ISSUE #505: Tabs: coupe of issues
Tabs should check for and handle `null`s in the child list, so you can do things like this:
```javascript
<Tabs>
  <Tabs.Tab label="Details">
    <DetailsTab {...props} />
  </Tabs.Tab>
  {props.enabled ? 
    <Tabs.Tab label="SSO">
      <SSOConfigTab {...props} />
    </Tabs.Tab>
    : null}
</Tabs>
```

Also, I think there's some other issues in the Tab component. It does not properly propagate `props` to itself and child tabs. For ex:
- https://github.com/auth0/cosmos/blob/master/src/components/molecules/tabs/index.js#L14
 if the label of a tab is set like this, you'll never get the update:
  `<Tabs.Tab label={props.resource.name}>`
- Same thing with children. Ex:
```javascript
<Tabs.Tab label={props.resource.name}>
   <DetailsView {...props} />
</Tabs.Tab>
```
If the props on DetailsView change, they are not reflected at all.

nkohari:
Fixes #505.


nkohari:
Closed in favor of #518.

PR #519, opened 2018-04-25
First implementation of Table component

nkohari:
This patch introduces a new `Table` component. As with `ResourceList`, `Table` is data-driven and will eventually support things like pagination, filtering, etc. However, `Table` is geared around rendering data on a cell-by-cell basis rather than on a row-by-row basis like `ResourceList`.

Here's an example. Let's say we have the following items that we want to display:

```js
const items = [
  {
    name: 'Isaac Asimov',
    born: 1920,
    died: 1992,
    image: 'data:image/jpeg;...'
  },
  {
    name: 'Ray Bradbury',
    born: 1920,
    died: 2012,
    image: 'data:image/jpeg;...'
  },
  {
    name: 'Philip K. Dick',
    born: 1928,
    died: 1982,
    image: 'data:image/jpeg;...'
  },
  {
    name: 'George Orwell',
    born: 1903,
    died: 1950,
    image: 'data:image/jpeg;...'
  }
]
```

Here's what a `Table` might look like:

```jsx
<Table items={items}>
  <Table.Column field="image" width="50px">
    {item => <Avatar image={item.image} />}
  </Table.Column>
  <Table.Column field="name" title="Name" width="40%" />
  <Table.Column field="born" title="Born" sortable sort="asc" />
  <Table.Column field="died" title="Died" sortable />
</Table>
```

As you can see, the children of `Table` are `Table.Column` elements. These elements aren't actually rendered; instead, they define how the individual cells should be rendered. By default, the item's value for the specified field is just rendered as text, but an optional render callback can be passed as a child of `Table.Column` to customize how it should be displayed.

This table would look like this when rendered:

screen shot 2018-04-25 at 1 26 15 pm

I think this can be merged as unstable as it is, but I'd like to get feedback on the API before we do too much more work on it. 😄 We also need to consider whether we want to target support for controlled tables, uncontrolled tables, or both. (That is, should the Table hold state about the current values for sorting, etc.?)

This will dovetail back into the conversation about how we want to support pagination and filtering for both `Table` and `ResourceList` as well.

Anyways, WDYT? 🤔 Feedback appreciated!


landitus:
The deployment link keep dying :( @siddharthkp may want to look at this at some point

siddharthkp:
This looks so polished! ✨

@landitus yeah, it's an annoying problem 😢 Open issue: https://github.com/siddharthkp/now-cd/issues/5

PR #520, opened 2018-04-25
Added additional pages to the PoC to be able to test components

landitus:
Added routing and stub pages for many sections of the PoC to test out components will real data and in a product context. This will also allow us to try mix and matching components in different situations. 

At first glance I think we can try and avoid having styled-components included in these pages so we can check the flexibility (or lack of) Cosmos components. Should we need to override something, we need to ack what's the proper way.

### DIFF
--- a/src/manage/components/page.js
+++ b/src/manage/components/page.js
@@ -4,6 +4,12 @@ import { BrowserRouter as Router, Route } from 'react-router-dom'
 import ClientIndex from '../pages/clients'
 import ClientDetail from '../pages/client-detail'
 import EmailsIndex from '../pages/emails'
+import RulesIndex from '../pages/rules'
+import SsoIndex from '../pages/sso-integrations'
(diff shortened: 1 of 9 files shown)


landitus:
Pending: Would be nice to have the sidebar reflect on the current page

nkohari:
This looks good, but I do think we should add the sidebar link highlighting before we merge it. I don't think it should be too difficult, but we'll need to shuffle around how our routes are working in the manage POC.

landitus:
@nkohari Do you think we could add a task to achieve the sidebar current page in the PoC? Maybe it's something we can do next. @siddharthkp may be able to help! 

I don't think it's Hi prio, we could merge this PR without it

nkohari:
Agreed, if the routing changes would take longer than a few hours of work it probably shouldn't hold up merging this PR.

PR #527, opened 2018-05-03
Button: Fix for the vertical aligment of text (check Firefox)

### LINKED ISSUE #528: Fix for the text vertical alignment in Buttons (check Firefox)
In Firefox, the text is misaligned inside the Button. I've managed to fix this with the `inline-flex` property, but I'm curious to check the Chromatic tests! 

screen shot 2018-05-02 at 21 35 16

This is a good opportunity to think about cross-browser testing (hopefully automated like Chromatic) ;)

landitus:
Fixes #528

### DIFF
--- a/src/components/atoms/button/index.js
+++ b/src/components/atoms/button/index.js
@@ -254,8 +254,9 @@ Button.Element = styled.button`
 Button.LinkElement = Button.Element.withComponent('a')
 
 Button.Text = styled.span`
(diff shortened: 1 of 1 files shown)


landitus:
That didn't go well 👎

nkohari:
After discussing this with @landitus, we're going to abandon this. The original bug still remains and needs attention, though.

Issue #525, opened 2018-04-30
Error when rendering Form Actions


landitus:
@ChrisWhiten [reported the following](https://auth0.slack.com/archives/C5ZK0DD8X/p1525104065000229):

> I'm getting a proptypes error whenever I render `<Form.Actions>`.  
Here's the error:
> > Failed prop type: Invalid prop `children` supplied to `ButtonGroup`

PR #529, opened 2018-05-03
Fix persistent warning in Form.Actions

siddharthkp:
Fixes https://github.com/auth0/cosmos/issues/525

Story time

`Form.Actions` uses `ButtonGroup`

`ButtonGroup` likes it's children to be a React element or an array of them

```js
ButtonGroup.propTypes = {
  children: PropTypes.oneOfType([
    PropTypes.element, // just one button for dynamic cases like Form Actions
    PropTypes.arrayOf(PropTypes.element)
  ])
}
```

Bad things happen when we have nothing to render, short version:

```jsx
<ButtonGroup>
	{ props.primaryAction && <Button/> }
	{ props.secondaryActions && <Button/> }
	{ props.destructiveActions && <Button/> }
</ButtonGroup>
```

In some cases, this can result in `children` to be `[React.element, undefined, undefined]` and it throws a warning because `undefined != PropTypes.element`

The ideal fix for me would be to replace undefined with [`React.Fragment`](https://reactjs.org/docs/fragments.html), but that's only available in React 16+

The other solutions are to 
- polyfill fragments
- make the prop type rules more flexible to accept `undefined`

```diff
- PropTypes.arrayOf(PropTypes.element)
+ PropTypes.array
```

I chose the latter.

### DIFF
--- a/src/components/molecules/button-group/index.js
+++ b/src/components/molecules/button-group/index.js
@@ -50,7 +50,7 @@ ButtonGroup.propTypes = {
   /** Should container only Buttons */
   children: PropTypes.oneOfType([
     PropTypes.element, // just one button for dynamic cases like Form Actions
-    PropTypes.arrayOf(PropTypes.element)
+    PropTypes.array
   ])
 }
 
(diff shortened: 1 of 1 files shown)


nkohari:
Hmm.. isn't `PropTypes.oneOfType([PropTypes.element, PropTypes.arrayOf(element)]` the same thing as `PropTypes.node`?

PR #530, opened 2018-05-07
support custom form fields

siddharthkp:


nkohari:
Is there a reason we can't do something like:

```jsx
<Form.Field label="Some label" helpText="Helper text">
  <MyInput placeholder="..." />
</Form.Field>
```

...and use `React.cloneElement()` to pass props like `error` on to the custom component? We're using this pattern in `Radio` now, and I think it results in a cleaner API.

siddharthkp:
Not really, that API would also work.

component felt slightly more natural to me because of how similar it is to our internal implementation

landitus:
Like the child element suggestion! Looks more simple (to me) just because it's more composable.

siddharthkp:
Would it be tricky to separate scope in the children example?

```jsx
const Container = (props) => (
  {props.label}
  <Form.Field label="Some label" helpText="Helper text">
    <MyInput placeholder={props.label} />
  </Form.Field>
)

render(<Container label="parent label"/>)
```

What would be the expected placeholder for the custom component?

landitus:
Sorry, I'm not following 😊

nkohari:
We probably just need to pass along a defined subset of props (eg. `error`) rather than all of them. Pretty much anything the custom component would need to render itself as a proper form field.

landitus:
> We probably just need to pass along a defined subset of props (eg. error) rather than all of them. Pretty much anything the custom component would need to render itself as a proper form field.

Seems like good a idea. Is there a downside to this?

siddharthkp:
Is it a smarter idea to not pass anything ourselves and let the user pass things explicitly? After all, it is their custom component.

```jsx
<Form.Field label="Some label" helpText="Helper text" error={this.state.errorMessage}>
  <MyInput placeholder={props.label} showErrorStyle={this.state.hasError}/>
</Form.Field>
)

```

nkohari:
Yep, I think you're right @siddharthkp. Better to just let them have control over it.

luisrudge:
What about using render props?

siddharthkp:
@luisrudge the first solution is a render prop! 😉

siddharthkp:
@nkohari Changed to support children

PR #532, opened 2018-05-07
Make file structure searchable

siddharthkp:
This PR runs the codemod from #531 

Before:

grop

After:

after

If the test cases pass, it means nothing broke 💪


siddharthkp:
That's weird because the builds passed 🤔

nkohari:
Do the docs work on your machine and/or is there anything else that might have changed that would cause it to break on mine? Master works fine for me, but this is broken. We should try to get this merged soon; since it touches so many things, it's already getting stale.

siddharthkp:
@nkohari Will back merge master into this

I'll clone the repository again to find what's the problem 🤔

PR #538, opened 2018-05-08
Reduce redundancy for applying Icon in Button

itaditya:
In current code, Icon and Button are being listed two times each. If we need to alter one thing in say the Icon then I need to make changes at two places. This is a pain. My PR intends to fix that

### DIFF
--- a/src/components/atoms/button/index.js
+++ b/src/components/atoms/button/index.js
@@ -158,21 +158,18 @@ const getAttributes = props => {
 }
 
 const ButtonContent = props => {
-  let content
-
-  if (props.icon && props.text) {
-    // The button contains both an icon and text.
-    content = [
-      <Icon key="icon" size={16} name={props.icon} color={getAttributes(props).icon} />,
-      <Button.Text key="text">{props.text}</Button.Text>
-    ]
-  } else if (props.icon && !props.text) {
-    // The button contains just an icon.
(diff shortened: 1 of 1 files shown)


src/components/atoms/button/index.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { Fragment } from 'react'
```
siddharthkp: We can't use fragments because we want to support React 15 as well 😄 (without polyfills if we can help it)

siddharthkp:
Thanks for taking the time!

src/components/atoms/button/index.js
itaditya: @siddharthkp, I did a little bit of googling and I found that the array syntax you are currently using is supported with React 16 onwards. See https://reactjs.org/blog/2017/09/26/react-v16.0.html#new-render-return-types-fragments-and-strings

The `Fragment` feature is added in React 16.2 .

itaditya:
Hey, what is the direction you are headed? Remove fragments to support React 15 or upgrade to react 16 ?

siddharthkp:
Still difficult to answer, but tending towards 16

nkohari:
Closing this for now since it's been open for awhile with no action. We can revisit this if needed.

PR #543, opened 2018-05-11
Make icon proptypes enum instead of string

siddharthkp:


nkohari:
This is a good idea. Instead of generating a new file, could we get away with just doing something like this in `icon.js`?

```js
import { icons } from './icons.json'

Icon.propTypes = {
  ...,
  name: PropTypes.oneOf(Object.keys(icons)),
  ...
}
```

siddharthkp:
@nkohari That doesn't work (yet) 😢

`react-docgen` parses that as `"value": "Object.keys(icons)"`, I'm trying to improve the setup so that we can at least do this:

```jsx
import icon, { iconNames } from '@cosmos/icon'

Button.propTypes = {
  icon: icon: PropTypes.oneOf(iconNames)
}
```

PR #544, opened 2018-05-14
Implemented Label component

nkohari:
Simple `Label` component:

screen shot 2018-05-14 at 4 27 19 pm

I wanted to confirm that I used the correct colors here; I sampled them out of the Sketch Cloud file, but I can't tell the actual color code that's being used. They seem very close to existing colors, but not an exact match. Either way I plan to standardize the colors between this component and `Badge` (which will be coming shortly).


nkohari:
Sorry for the force-push; wanted to bring this up-to-date with the new `index.js` restructuring without a merge commit.

PR #547, opened 2018-05-15
Make file structure searchable (attempt 2)

siddharthkp:
Rerunning https://github.com/auth0/cosmos/pull/532

Before:

grop

After:

after

If the test cases pass, it means nothing broke 💪


siddharthkp:
@nkohari Fixed the docs and ran the codemod on master

Issue #509, opened 2018-04-16
Provide an image-only Avatar component


nkohari:
Right now, the `Avatar` component has branching logic which will cause it to render differently if a `name` prop is specified. However, we want to be able to use the `Avatar` as an atom to build other components, and sometimes the alternative render mode won't be compatible (as in the case of the `ResourceListItem`).

There are two options:

1. Make the `Avatar` component just render the image (similar to `Thumbnail`), and create an `AvatarBlock` component that will render the image (using `Avatar`) along with the user's name.
2. Make it so `Avatar` requires a `name` component, and add a `circular` prop to `Thumbnail` to allow it to act as the image-only version.

These have pros and cons; for example, option #2 means we can remove some redundancy in the styling between the two components, but it may be more confusing to users. Likewise, it may not be immediately clear when a user should choose an `AvatarBlock` instead of an `Avatar`.

Both options will introduce a breaking change, so we should consider carefully before changing anything. 😄

nkohari:
Closed in favor of #561.

Issue #406, opened 2018-03-14
Avatar spec needed


siddharthkp:
I don't know what's expected from the `Avatar` component, how many types of avatars exist, what design aspects should be locked and what should be cusomizable (background-color, shape, size, etc.)

There is some discussion here: https://github.com/auth0/cosmos/pull/398

landitus:
I did a small inventory for the **Avatar**: 

<img width="751" alt="screen shot 2018-03-14 at 17 02 35" src="https://user-images.githubusercontent.com/239215/37428748-fcd552a0-27ab-11e8-8d81-f80452fb0c90.png">

### The User Avatar:
The default user avatar has a letter or 2, and a random background color. The color stays with you, so it's not random for each rendering. I was talking to @vctrfrnndz and he was telling me that this default lettering system is actually pulling images from the CDN will all possible single and double letter combinations with a random color. This is because we are using gravatar and the fallback should be an image (not a text inside a circle). Maybe we need to dig deeper into this. 🧐

If the user has an image, it will render the photo to fill the circle.

### The Image for an entity:
For an API or Client or Connection. It's a full image with no background. It has rounded corners and a border. I would expect the image fill the square. We should have a background color of gray as a fallback, for loading purposes.

There is a special case where an icon (from our library) is used inside the square to represent an entity. We should be able to achieve this too.

### Ideas
Have 2 components: `Avatar` and `Thumbnail`. Move forward with the Thumbnail and figure out the Avatar later?

landitus:
Sometimes this inventory / discovery can lead up to think or propose a change in the **design**  to condense disconnected UI design decisions. I wouldn't discard this, as when we do componentization work, it's useful to do some cleanup.

beneliflo:
`Thumbnail` love it. That is perfect name to this component and it's easier to identify it and differentiate it with the user avatar.

landitus:
In this case, the thumbnail has an outline (gray) but inside can be either an image or an icon. So, when we want to show a logo for inside a table, we should use the Thumbnail to add the image.

For example, here:
<img width="901" alt="screen shot 2018-03-16 at 19 07 38" src="https://user-images.githubusercontent.com/239215/37546784-55a84cd2-294d-11e8-9613-e38aa61c2fbd.png">

landitus:
Love if the avatar could handle the Gravatar logic! We should first check how it's done today

astanciu:
Wanted to add another Avatar to the inventory:
[...]

PR #561, opened 2018-05-24
Add AvatarBlock component

nkohari:
Originally this was spec'd as creating a new component called `AvatarBlock`, which would support both a circular image (used primarily for user profile pictures) as well as a square containing an icon (used primarily for resources). The goal was to have `AvatarBlock` consume `Avatar` (for user images) and `Thumbnail` (for icons, primarily).

As I started working through this, I realized that while this structure would result in several small, very focused components, the code required for `AvatarBlock` would be very complex -- and worse, the API would be very cumbersome to use. As a result, I decided to take a stab at rolling all of the functionality into a single `Avatar` component, which would replace `Avatar`, `AvatarBlock`, and `Thumbnail`.

screen shot 2018-05-24 at 3 15 23 pm
screen shot 2018-05-24 at 3 15 26 pm

The main place we were using `Thumbnail` was in `ResourceListItem`, which I replaced with `Avatar` ... but I realized that the "header" component of a list item looks almost identical to an `Avatar`, and even has the same props. Although it may not be a good idea to couple the list item to the avatar, we may consider replacing the list item header entirely with an `Avatar`, passing on the `title` and `subtitle`, etc. props from `ResourceListItem`.


nkohari:
In #406, I mentioned the idea of building Gravatar support into `Avatar` by use of an `email` prop. It wouldn't be too tough to add this here if that's something we want to support.

nkohari:
If this is accepted, we can also close #509 in favor of this.

nkohari:
@siddharthkp Mind taking a look at this when you have a chance?

nkohari:
Sounds good! I think the important piece of the work here is merging `Avatar` and `Thumbnail`. From there, creating `AvatarBlock` is much more straightforward, since it can't just accept an avatar with an `avatar` prop.

nkohari:
Adding the `don't merge` label so I can implement the changes we talked about this morning.

nkohari:
Alright! This patch has been updated to add the `AvatarBlock` component. We're still combining `Avatar` and `Thumbnail`, but the title and subtitle are split off into another component as was originally spec'd.

`AvatarBlock` supports 3 sizes (`compact`, `default`, and `large`), which map onto `Avatar` sizes of `small`, `medium`, and `large`.

The `shape` property of `Avatar` has been removed, and replaced with a `type` prop, which can accept values of `user` or `resource`. The former results in a circular image (like the old `Avatar`) and the latter results in a square image with a gray background (like the old `Thumbnail`).

I also rebased this patch against the current master to fix the collisions; sorry for the force push. 😄

landitus:
I'll take a look ! Just noticed the following bug in the docs when using the props table:

kapture 2018-06-12 at 12 56 38

landitus:
The Client Page Header and the Resource List: Clients examples in the Overview page and in the PoC: manage/clients and manage/clients/apXVn76xBT7WougWj17MGXHSthfrBB have broken avatars. Might be out of date.

landitus:
Found that some components are breaking (due to these changes): 

- PoC Local components: Clients List and Client detail page
- Overview page

screen shot 2018-06-18 at 15 55 10
screen shot 2018-06-18 at 13 08 40

PR #563, opened 2018-05-28
Add sketch files for atoms + molecules

siddharthkp:
Here's the library file to play around with: [Dropbox link](https://www.dropbox.com/s/bohw74sp46qiyh4/cosmos-lib.sketch?dl=0)

Problems:

1. What should be the default width of the container? 800px? 1024px?

1. Some components that depend on position don't render very well like Switch and Tooltip

	screen shot 2018-05-28 at 4 25 46 pm

1. Inputs, Selects, etc. don't get placeholders


siddharthkp:
```ci/chromatic — 99 specs unchanged``` 

**phew**

PR #576, opened 2018-05-31
Add support for left and right positions to Tooltip, code cleanup

nkohari:
This adds support for the `left` and `right` values to the `position` prop of `Tooltip`.

screen shot 2018-05-31 at 2 51 42 pm

We should consider whether we want to change from `defaultVisible` and the `&:hover` class to a `visible` prop to support a controlled version of the component.

### DIFF
--- a/src/components/atoms/tooltip/tooltip.js
+++ b/src/components/atoms/tooltip/tooltip.js
@@ -4,55 +4,93 @@ import PropTypes from 'prop-types'
 
 import { colors, spacing, misc } from '@auth0/cosmos-tokens'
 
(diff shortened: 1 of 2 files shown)


siddharthkp:
I don't think this should be a controlled component, the expectation from a tooltip is usually to appear automatically on hover

nkohari:
In Slack, we discussed changing `defaultVisible` to `visible` with the possible values of `true`, `false`, and `auto`. For now I'm just going to merge this as-is, but we may want to add that later to solve the controlled/uncontrolled question.

Issue #580, opened 2018-06-02
Radio.selected not working


astanciu:
This line fails the check even when the child is RadioOption
https://github.com/auth0/cosmos/blob/master/src/components/atoms/radio/radio.js#L111

Using a debugger...
![image](https://user-images.githubusercontent.com/1675025/40879342-58c368b4-666c-11e8-94f5-dbbbaeb1526e.png)

Maybe it's related to how Webpack bundles them??

Issue #578, opened 2018-06-01
Syntax error in documentation for Form Fields


landitus:
There is a syntax error in the docs: https://auth0-cosmos.now.sh/docs/form#form-fields

![screenshot-auth0-cosmos now sh-2018 06 01-18-10-31](https://user-images.githubusercontent.com/239215/40863716-36a66628-65c7-11e8-8c07-6d56641c6ef5.png)

PR #581, opened 2018-06-04
Fix radio bugs

siddharthkp:
- Fixes https://github.com/auth0/cosmos/issues/580

	`child.type === RadioOption` only works in dev, because `RadioOption` gets compiled to `React.createElement`. This check seems to be a validation on the children and we probably don't need it there. @beneliflo @nkohari can confirm?

- Fixes https://github.com/auth0/cosmos/issues/578

	fixed syntax error


siddharthkp:
> Yep, this makes sense

Cool, I'll remove that in another PR

PR #585, opened 2018-06-04
Tabs: Add tab controlled state mode

### LINKED ISSUE #579: Tabs: needs onSelect()
Chosen tab is lost in a complex component when the state changes and a render happens. Need to be able to keep track of which tab is selected and re-select it after a state change. 

For ex:

```JSX
const setTab = (index) => this.selectedTab = index;

<Tabs onSelect={setTab} selected={this.selectedTab}>
  <Tabs.Tab />
  <Tabs.Tab />
</Tabs>
```

Currently, you can set selected on `Tabs.Tab` but there's no way to programmatically set that based on some condition.

```JSX
<Tabs.Tab selected={what do I have available here to compare?} />
```

francocorreasosa:
This PR enables a controlled state mode in `Tab`. It will enable the users to enforce a selected index by passing a `selected` prop and subscribe to tab change intents from the user with the `onSelect` prop.

Resolves #579.

### DIFF
--- a/src/components/molecules/tabs/tabs.js
+++ b/src/components/molecules/tabs/tabs.js
@@ -2,6 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import { colors, spacing } from '@auth0/cosmos-tokens'
+import makeId from '../../_helpers/uniqueId'
 
 const Wrapper = styled.div``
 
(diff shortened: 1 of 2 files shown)


siddharthkp:
Hey @francocorreasosa, I've added an example in the docs to show what's broken. I don't think it works with the code changes you've made

francocorreasosa:
Added the `Don't merge` label so you can test the use case of the bug and then I can remove it before merging.

landitus:
Can you take a look @siddharthkp ?

PR #590, opened 2018-06-05
Page Header: show secondary action when primary is missing.

### LINKED ISSUE #589: Page Header fails to render only one action button
@jcenturion reports: 
https://github.com/auth0/cosmos/blob/master/src/components/molecules/page-header/page-header.js#L27-L51 => If I do something like:

```
<PageHeader
  title="Applications"
  description={{
    text: 'Setup a mobile, web or IoT application to use Auth0 for Authentication.'
  }}
  primaryAction={null}
  secondaryAction={{
    label: 'Tutorial',
    icon: 'play-circle',
    method: () => {}
  }}
/>
```

I ended up losing both buttons.

*Use case*: Based on a feature flag. I should display or hide a button. In this case ‘Create application’.

francocorreasosa:
We now display the secondary action even if the primary one is missing.

Added two new stories:
- just renders the primary button in the Page Header to test this use case
- just renders the secondary button in the Page Header to test this use case

Resolves #589

### DIFF
--- a/src/components/molecules/page-header/page-header.md
+++ b/src/components/molecules/page-header/page-header.md
@@ -25,7 +25,9 @@
 />
 ```
 
-## Single actions
+## Examples
+
+### Single actions
 
 A Page header can have a single primary action:
 
@@ -44,7 +46,7 @@ A Page header can have a single primary action:
 />
 ```
 
-A Page header can have a single secondary action:
+A Page header can also have a single secondary action:
 
 ```js
 <PageHeader
--- a/src/components/molecules/page-header/page-header.md
+++ b/src/components/molecules/page-header/page-header.md
@@ -24,3 +24,33 @@
   }}
 />
 ```
+
+```js
+<PageHeader
+  title="Clients"
+  description={{
+    text: 'Setup a mobile, web or IoT application to use Auth0 for Authentication.',
+    learnMore: '/clients'
+  }}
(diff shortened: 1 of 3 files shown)


src/components/molecules/page-header/page-header.md
```diff
@@ -24,3 +24,33 @@
   }}
 />
 ```
+
```
landitus: Let's add an **Example** title here, and then the following:

### Single actions
A Page Header can have a single primary action:

src/components/molecules/page-header/page-header.md
```diff
@@ -24,3 +24,33 @@
...
+    learnMore: '/clients'
+  }}
+  primaryAction={{
+    label: 'Create Client',
+    icon: 'plus',
+    handler: () => {}
+  }}
+/>
+```
+
```
landitus: Let's add the following paragraph:

A Page Header can have a single secondary action:

src/components/molecules/page-header/page-header.md
francocorreasosa: Done!

src/components/molecules/page-header/page-header.md
francocorreasosa: Done!

landitus:
LGTM!

Issue #592, opened 2018-06-06
Docs search is broken


francocorreasosa:
As I demo in the GIF, as soon as I type a letter in the search box (`s` in this case), the entire site will break.

The site won't always break. Only when starting with `e`, `t`, `i`, `s`, `d`, `l`, `n`, `m`, `I`, `A`.

![kapture 2018-06-06 at 9 08 42](https://user-images.githubusercontent.com/4152942/41037349-77f0da1e-6969-11e8-8378-615991f00785.gif)

Issue #480, opened 2018-04-08
Docs page searching feature breaks the app when I type certain characters


m-sureshraj:
When I **start** to type certain characters e.g. (i, t, c) etc. the whole application get's broken.

![cosmos searching issue](https://user-images.githubusercontent.com/12813750/38463388-7df6555e-3b17-11e8-9130-6076f9aa446b.gif)

**Steps to reproduce the issue**
* Go to [https://auth0-cosmos.now.sh/docs](https://auth0-cosmos.now.sh/docs) page
* Type 'c | t | i' character in the search input
* Inspect the console

**I tried to find the issue but I couldn't set up this repo properly.** 
* After installing the dependencies ```yarn dev``` cmd could not resolve the ```@auth0/cosmos``` package. Why this package not added to the dependencies list?

<img width="1049" alt="screen shot 2018-04-08 at 10 45 22 am" src="https://user-images.githubusercontent.com/12813750/38463495-f9378312-3b19-11e8-8f30-c0f7b8a61c17.png">

* I manually install the ```@auth0/cosmos``` package & tried to build the dev again, got following error.
 
<img width="1064" alt="screen shot 2018-04-08 at 10 55 16 am" src="https://user-images.githubusercontent.com/12813750/38463562-9d7a3c02-3b1b-11e8-810d-f8855a30571b.png">

I really like to help to fix this issue but please help me to setup this repo.

Thanks.

nkohari:
Closed by #593.

PR #593, opened 2018-06-06
Avoid breaking docs site on specific searches

francocorreasosa:
Resolves https://github.com/auth0/cosmos/issues/592

### DIFF
--- a/src/docs/sidebar/children.js
+++ b/src/docs/sidebar/children.js
@@ -17,7 +17,7 @@ const attachChildren = components => {
 
   children.forEach(child => {
     const parent = getParent(parents, child)
-    parent.children.push(child)
+    if (parent) parent.children.push(child)
   })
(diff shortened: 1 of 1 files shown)


landitus:
Added a ref to the issue

landitus:
Does this solve https://github.com/auth0/cosmos/issues/480 too? they loom quite similar

nkohari:
Yep! Good catch.

src/docs/sidebar/children.js
```diff
@@ -17,7 +17,7 @@ const attachChildren = components => {
 
   children.forEach(child => {
     const parent = getParent(parents, child)
-    parent.children.push(child)
+    if (parent) parent.children.push(child)
```
siddharthkp: 🙈👍👍

PR #603, opened 2018-06-07
Fix button vertical spacing in Firefox

francocorreasosa:
Just removes `vertical-align: middle;` from the button text.
 
firefox button vertical align


landitus:
Good, it fixes the bug for FF, but introduces a slighly very tiny visual regression back in Chrome. So, now in Chrome, the text is offcentered by 1px or 2px. We where thinking to let @beneliflo take over and see what's the root of this problem.

PR #575, opened 2018-05-31
Pull manage out of src

siddharthkp:
There's a fundamental problem in our development flow, I think we've talked about it before

How we build cosmos != how others consume it

Bugs like https://github.com/auth0/cosmos/issues/570 happen 

1. because our dev environment is tightly coupled, our PoC doesn't have it's own build process
2. we don't have tests for a built version 🤔 

Addressing problem 1 in this PR without disturbing our dev environment _by much_

----

Update: 

- For staging and production, there's basically no difference. (there's a cute little hack in `public/redirect`)

- For local development of manage PoC you need to go to `examples/manage` and run `yarn dev` there (it's no longer in the same app as src).
	It's hooked up to yarn workspaces so any changes made to cosmos components will propagate as usual

- Can't share components between difference example apps. Example: overview cannot import components from manage

PR #607, opened 2018-06-11
Pull docs to internal

siddharthkp:
work in progress

depends on https://github.com/auth0/cosmos/pull/575 for `/redirect`


nkohari:
Is this ready to go now that #575 has been merged?

siddharthkp:
@nkohari We're good now!

landitus:
Would it make sense to have the docs in `/docs` instead of `internal/docs` ?

siddharthkp:
File structure: There's more to come in internal like overview, sandbox, etc.

URL: It's still accessible at the same url of auth0-cosmos.now.sh/docs


Here's the plan:

map

landitus:
perfect!

PR #610, opened 2018-06-12
Document how to use styled + extending

siddharthkp:
PR will change after https://github.com/auth0/cosmos/pull/607

This is just for looking at the copy: https://auth0-cosmos-document-styled.now.sh/docs/usage

### DIFF
--- a/src/docs/home.js
+++ b/src/docs/home.js
@@ -22,7 +22,7 @@ class Home extends React.Component {
             <Pre>npm install --save @auth0/cosmos</Pre>
             If you prefer yarn, run this command instead:
             <Pre>yarn add @auth0/cosmos</Pre>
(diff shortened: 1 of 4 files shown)


siddharthkp:
That's true. If given a change we would like folks to not extend it. 

But I guess, this is an escape hatch in the odd case they do?

landitus:
A Style override that changes appearance, like color or size should not allowed as all allowed options are what the system has in the documentation. 

If the case a style doesn’t work or is troublesome, a request for a change will certainly let us have more info and that will also push changes to the library to all our users. An override just patches it and hides the problem and the possibility to solve it.

Maybe there are other use cases that we can discuss! I want to exhaust all options in which we can fix on our end first. 💪

siddharthkp:
This PR was only for preview (does not need to be open)

PR #612, opened 2018-06-12
Checkbox component

beneliflo:


siddharthkp:
Yep, I agree. It took some time to understand the usage in the docs

I really think making them individual instead of a group will simplify the work you have to do in `onChange`

landitus:
good points ! I haven't realized that distinction, so I agree on giving a shot on improving it.

PR #614, opened 2018-06-13
Export styled from cosmos

### LINKED ISSUE #596: Using a styled() override breaks CSS styles
(Reported by @jcenturion)

When adding the following lines from the [client list example](https://github.com/auth0/cosmos/blob/poc-aditional-pages/src/manage/components/client-list.js):

```js
const ClientID = styled(StyledTextSubdued)`
  margin-right: ${spacing.xsmall};
  line-height: normal;
`
```

... it causes styles to disappear:

image

I haven't tried to reproduce the error, but we should look into it and see what might be causing it.

francocorreasosa:
Resolves #596


src/components/styled.js
```diff
@@ -0,0 +1,3 @@
+import styled from 'styled-components'
+
+export default styled
```
siddharthkp: We should also export all the helpers that come from styled-components like `injectGlobal`, `keyframes`, `css`, etc. (https://www.styled-components.com/docs/api#helpers)

PR #620, opened 2018-06-18
Add Dialog example in documentation and storybook.

### LINKED ISSUE #510: Display actual Modal and Dialog components in documentation instead of just screenshots
Today, we don't have a good way of displaying the Dialog in the documentation. One quick fix would be to add a button with an Open Dialog label that triggers the opening of the docs, similar to what we have in the overview page.

francocorreasosa:
Resolves #510


landitus:
Nice! I believe this can be improved if we move the example code block to the main spot. Even though it has more code related to the state, it's the best example and representation of the component.

screen shot 2018-06-19 at 09 11 30

PR #624, opened 2018-06-19
Implement form context (Fixes #619)

### LINKED ISSUE #619: Form layout is not passed down to children
Reported by @jstrutz here: https://github.com/auth0/cosmos/issues/177#issuecomment-397385427

A couple notes on the things we've found, when working w/ forms:

### Surrounding a `<Form.TextInput>` element (and probably other types) with a non-presentational component breaks the form layout
```jsx
<Form layout="label-on-left">
   <>{/* Note the wrapping Fragment */}
   <Form.TextInput
     label="Email"
     name="email"
     type="text"
     placeholder="Your email address"
  />
  </>
</Form>
```
screen shot 2018-06-14 at 2 58 01 pm

### Action Handlers need disabled state

So we can communicate that the form cannot be submitted (i.e. in the case of invalid input).  Example:

```jsx
<Form.Actions primaryAction={{
  disabled: true,
  label: 'Save Changes',
  handler: handleSubmit
}} />
```

siddharthkp:
- Fixes https://github.com/auth0/cosmos/issues/619

- Upgraded peer dependency to 16.3 because now we use context API


landitus:
Is there some black magic going on? 👹

src/components/molecules/form/form-context.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react'
+
+const LayoutContext = React.createContext()
```
nkohari: If we only have a single context for forms, won't we have problems if a user puts two forms with different layouts on the same page?

nkohari: Ah, I missed this part in the 16.3 docs:

> When React renders a context Consumer, it will read the current context value from the closest matching Provider above it in the tree.

Based on this, we should be fine, right? 😄

PR #627, opened 2018-06-19
Fix button minimum heights (also affecting icon vertical align)

### LINKED ISSUE #625: Vertical alignment/height of some embedded components is wrong
We discussed this in #575. At some point a bug was introduced involving the vertical alignment/height of some embedded components. Here are some examples:

It seems like it might just be an issue with `line-height`?

francocorreasosa:
Resolves #625.

### DIFF
--- a/src/components/atoms/button/button.js
+++ b/src/components/atoms/button/button.js
@@ -201,6 +201,7 @@ const Button = ({ children, ...props }) => {
 
 Button.Element = styled.button`
   display: inline-block;
+  min-height: ${props => getAttributes(props).lineHeight};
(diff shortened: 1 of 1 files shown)


siddharthkp:
Great find!

Makes me wonder is there a bug in our height setting logic? (that is based on props)

(We can go ahead and merge this PR either way and then come back to a little cleanup)

landitus:
Looks like a very simple fix @francocorreasosa !  I'm still intrigued, as it looks the line-height is not being applied to the button (but the min-height is). As @siddharthkp suggested, we can merge this and open a new PR to get into the real issue of what's going on.

PR #628, opened 2018-06-19
Adding support for TSD

jcenturion:
The goal of this PR is to add support for Visual Studio Code IntelliSense by adding [TypeScript declarations](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html). 

## Updated publish script

image

## Demo

tsd-cosmos

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,5 @@ dist
 # ignore generated files
 src/components/meta
 internal/docs/public/main.js
+
+*.d.ts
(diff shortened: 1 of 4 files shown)


siddharthkp:
@nkohari Need your help in reviewing this. I don't understand what I'm looking at 😅

nkohari:
This is great! I'm wondering if it makes sense to have it in a different part of the build process, though. Having it in the publish step means we won't get the benefit of the definitions during development, and as a VS Code user that makes me sad. :)

nkohari:
Closing this in favor of #632, which is really just this work moved to a different step in the build. Thanks @jcenturion, this is great!

PR #634, opened 2018-06-22
Fix button align

beneliflo:
After a research the height of the buttons must be odd to have consistency in the cross browsing.


nkohari:
Closed in favor of #635.

PR #635, opened 2018-06-22
Fix button alignment (via @beneliflo)

nkohari:
Supercedes #634. Re-created against current master to avoid a problematic merge.

I'm not sure if this achieves what we're going for; I'm just shuffling the code around. 😄

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "if-env": "1.0.4",
     "jscodeshift": "0.5.0",
     "latest-version": "3.1.0",
-    "lint-staged": "7.0.0",
+    "lint-staged": "^7.0.5",
     "lodash.camelcase": "4.3.0",
     "lodash.frompairs": "4.0.1",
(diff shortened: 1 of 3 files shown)


landitus:
@beneliflo you can go to https://www.chromaticqa.com/build?appId=5aba7f6a07f95800202658ff&number=383 and then go to each item to see the snapshots comparison. The easiest way to check is to use the toolbar and choose 1-up and then toggle the switch. 

Here's an example:
kapture 2018-06-22 at 14 14 23

landitus:
I'm adding the buttons design specification so it's easier to compare. 

screen shot 2018-06-22 at 14 18 39

Note that the **button height** is also related to the **input height**, in cases when the elements are next to each other, like the example below (from github, but we could do something similar)

screen shot 2018-06-22 at 14 07 52

nkohari:
Closing this due to inactivity.

PR #636, opened 2018-06-22
Rename clients to applications in manage PoC.

francocorreasosa:


nkohari:
Blocked by #650.

nkohari:
This may end up causing conflicts with #650 but I'm going to merge it now, and resolve any problems we run into later.

PR #639, opened 2018-06-25
Pull overview out

siddharthkp:
Collateral: sketch sandbox doesn't work anymore. Can merge without it though. 

Will rewire it next when sandbox get's a place of it's own

Chromatic tests passed.


siddharthkp:
Merged as a part of #645

PR #640, opened 2018-06-25
Delete old screenshot diffing

siddharthkp:
We don't use this anymore


siddharthkp:
Merged as a part of #645

PR #641, opened 2018-06-25
Rename src to core

siddharthkp:
Rename src to core so that folks don't confuse that as the starting point

map


siddharthkp:
We don't even need this after #645

PR #645, opened 2018-06-26
Engine 0.1

### LINKED ISSUE #651: Local apps seem to use output of yarn build
I was reviewing the new `Checkbox` component, and when running `yarn scripts docs.dev`, it didn't appear in the sidebar of the docs site. After restarting the app a couple of times with no change in behavior, I tried running `yarn build`, after which the `Checkbox` appeared on the sidebar.

Is it possible the local apps are using the production bundle output somehow? It seems like this is similar to the issue we ran into the other day which required removing the `main.js` file from our local copies.

siddharthkp:
Built on top of https://github.com/auth0/cosmos/pull/639 and https://github.com/auth0/cosmos/pull/641

For a nicer diff, check this out: https://github.com/auth0/cosmos/compare/temp-branch...engine-0.1

Fixes #651 and #652

`engine 0.1`

### What this is:

Consistent tooling for applications that use cosmos. It's for us to use in example apps, sandbox, etc.  This should make our lives easier.

When it's nicer to use, we'll also pass this on to applications outside the cosmos repo

### What this means for applications:

No more config

screen shot 2018-06-28 at 8 18 21 pm

Simplified tooling

screen shot 2018-06-28 at 8 18 46 pm

(We had this with create-react-app, but not with the flexibility)


landitus:
What's all this black magic about? 🧐

siddharthkp:
@landitus Updated the description + changed the base so the diff looks better 😄

landitus:
Much better! jajaj

siddharthkp:
Well, I tried an isolated environment with `examples/hot-world` and that works perfectly.

But, when I use the same setup in `webpack-hello-world` from the `engine`, it fails without errors 😢 

The log on the console stops at `App hot update...`, doesn't actually update anything

```
[HMR] Waiting for update signal from WDS...
client:71 [WDS] Hot Module Replacement enabled.

[WDS] App updated. Recompiling...
client:217 [WDS] App hot update...
```

@nkohari @francocorreasosa Have you folks seen this before?

siddharthkp:
Works well here: https://github.com/siddharthkp/hmr-test

Which means it's probably a dependency mismatch problem because of all the different webpacks roaming around?

nkohari:
@siddharthkp It looks like webpack is pinned to `4.8.3` in the manage example app. Is that the version you're running? The other libraries (`webpack-cli` and `webpack-serve`) are also pinned to a specific version.

siddharthkp:
Sometimes you have to break everything apart to make it work again

- [] core/components
- [] core/tokens
- [] core/codemods
- [x] core/babel-preset
- [x] core/cli
- [] internal/overview
- [] internal/docs
- [] examples/manage
- [x] examples/webpack-hello-worl

siddharthkp:
@nkohari I think we're good now!

nkohari:
I'm still seeing the same problem with HMR, but this patch is definitely N+1 from where we are on master right now. I'm going to merge it as-is but we should see if other folks have trouble with HMR, or if it's just me. 😄 Maybe we should compare environments to see if it's something configuration-related?

siddharthkp:
Yep, can confirm. It's broken again 😭

siddharthkp:
Unrelated: Found that `nps-i` eats up the logs somewhere.

The logging experience of using `yarn scripts manage.dev` and `yarn i > manage.dev` is different. I'll check what's wrong in `nps-i`

siddharthkp:
@nkohari Stories also work now 😄

PR #650, opened 2018-06-29
Avatar now accepts an <svg> for its image prop

nkohari:
This fixes an issue with the client list in the Manage example app, but is also a nice-to-have feature.

Technically it will support any component right now, but I didn't want to start doing checks since we've run into problems before with that and bundling.


landitus:
We still need to fix the size prop for the Avatar in this local component 

screen shot 2018-06-29 at 12 09 45


It's a page header inside the PoC

nkohari:
@landitus Where are we with being able to merge this?

PR #660, opened 2018-07-04
#658: Support raw buttons as dialog action attributes.

### LINKED ISSUE #658: Dialog.Action does not support al button attributes
**Use case**

As a developer, I would like to be able to use other Cosmos Button attributes like `success` and `loading` to send feedback to the user about what is happening during an async call.

**Posible solution**

A possible solution could be to just support Cosmos Buttons in `actions`. For example:

```jsx
    <Dialog
      open
      title="Example Dialog"
      onClose={() => this.setDialogState(false)}
      actions={[
        <Button loading onClick={() => { ... }}>Save</Button>
      ]}
    >
      Are you sure?
    </Dialog>
```

**Live code**

https://codesandbox.io/s/9zz6r5oo7p

**Sample code**

```jsx
    <Dialog
      open
      title="Example Dialog"
      onClose={() => this.setDialogState(false)}
      actions={[
        new Dialog.Action(
          'OK',
          () => {
            alert("You've performed the 'OK' action.")
          },
          'primary'
        ),
        new Dialog.Action('Cancel', () => {
          alert("You've performed the 'Cancel' action.")
        })
      ]}
    >
      Are you sure?
        </Dialog>
  );
```

francocorreasosa:
Resolves #658


nkohari:
@siddharthkp What do you think about this approach?

siddharthkp:
I like it, it's more flexible and future proof

How do you feel about making this the default way (and deprecating the other)

nkohari:
I think having both as an option is reasonable. "Easy" and "advanced" mode maybe? I'm ok with either approach though.

nkohari:
Either way I'm going to merge this as-is. We can decide to take another step with it in a later PR if we choose to do so.

Issue #644, opened 2018-06-26
Linking in documentation keeps the scrolling position between pages


landitus:
Linking between pages in the documentation persists the scrolling position, which it's not the expected behavior.

![kapture 2018-06-26 at 8 04 16](https://user-images.githubusercontent.com/239215/41907660-ef9a0cc6-7917-11e8-81ae-89a6410796db.gif)

landitus:
I still see have the same problem. Re-opening this :)

PR #663, opened 2018-07-05
Docs: Scroll to top on page change

siddharthkp:
Based on https://github.com/auth0/cosmos/pull/645

Fixes https://github.com/auth0/cosmos/issues/644


landitus:
This is definitely better! but still doesn't load at the top of the page as expected. There is a margin at the top of the Page title that is cut off.

screen shot 2018-07-06 at 06 50 50

PR #666, opened 2018-07-05
WIP: Pagination component base

francocorreasosa:
Had to create a PR to get a now.sh URL to share. Labeling as `don't merge` for the time being.

### Pagination
imagen

### Pagination Toolbar
image

### Pager
image


core/components/molecules/pagination/pagination-toolbar.md
```diff
@@ -0,0 +1,71 @@
+```meta
```
siddharthkp: I don't think we need to document `pagination-toolbar`, it's an internal component 

(p.s. it also breaks on the docs right now)

core/components/molecules/pagination/pagination.md
```diff
@@ -0,0 +1,36 @@
+```meta
+  category: Layout
+  description: "Paginate resources with a handy pagination component"
+```
+
+```jsx
+<Pagination items={20372} perPage={10} page={3} />
```
siddharthkp: To make the prop table work, you need to pass `{props}` to the component and you can set the intended defaults with the `defaults` prop

Example:

```jsx
< Pagination {props} defaults={{items: 20372, perPage: 10, page: 3}} />
```


More here: https://auth0-cosmos-master.now.sh/docs/#/contribution-guide

core/components/molecules/pagination/utils.js
```diff
@@ -0,0 +1,86 @@
+/**
+ * Calculates the page input width based on the characters
+ * it contains.
+ * @param {number} page
+ */
```
siddharthkp: yay code comments! 💯

core/components/molecules/pagination/pagination-toolbar.md
francocorreasosa: Yep I know. Working on that, so should we completely remove that component from docs?

landitus: > I don't think we need to document pagination-toolbar, it's an internal component

We need to show how to use this component somehow. The easiest way, for now, is to move it to our components list (not inside pagination, but as a standalone component).

siddharthkp: Ah, you're right.

I assumed wrongly that `pagination-toolbar` is a child of `pagination`, it's a separate compound component

francocorreasosa:
@siddharthkp @landitus Can you please take a look again to check if I'm missing something?

core/components/molecules/pagination-toolbar/pagination-toolbar.md
```diff
@@ -0,0 +1,71 @@
...
+    )
+  }
+}
+```
+
+### Example with totals
+
+You may want to show a little bit more detail of the content you are
+displaying, so we let you show the totals for the current page by passing the `showTotals`
+prop to the `<PaginationToolbar>` component.
```
siddharthkp: Not sure if that's the best name, but I can't think of a better one either 😅

`showIndex`,` showLegend`

@landitus how would you describe this component?

core/components/molecules/pager/pager.js
```diff
@@ -0,0 +1,47 @@
...
+      Older {'>'}
+    </Button>
+  </StyledPager>
+)
+
+Pager.propTypes = {
+  page: PropTypes.number.isRequired,
+  perPage: PropTypes.number.isRequired,
+  items: PropTypes.number.isRequired,
+  onPageChanged: PropTypes.func
```
siddharthkp: As an action, this should be `onPageChange` (drop the d at the end)

core/components/molecules/pager/pager.js
```diff
@@ -0,0 +1,47 @@
...
+`
+
+const Pager = ({ onPageChanged, page, perPage, items }) => (
+  <StyledPager>
+    <Button
+      size="small"
+      appearance="secondary"
+      onClick={() => changePageIfAppropiate(page - 1, items, perPage, onPageChanged)}
+    >
+      {'<'} Newer
```
siddharthkp: We should replace the arrows with icons. You can achieve this with `chevron-down` + rotation if you feel like it

@landitus request for `chevron-left` + `chevron-right`

siddharthkp:
The vertical alignment is slightly off

image

siddharthkp:
This PR has a lot of good things, we should merge this and make the mentioned improvements in a different pull request

siddharthkp:
Feature requests:

Can we merge the 3 styles of pagination introduced into this pull request as one component `[ Pager, Pagination Toolbar, Pagination ]`

For a person looking for a pagination component in cosmos, it would be confusing to pick the right one because they all have similar names (`Pager` and `Pagination` have the same description as well 😅)

Suggestion:

- In case of specialised components like `PaginationToolbar` which are built for tables, we should attach them to the table (`Table.Pagination`) instead of making them available as a new component.

- For components that are too similar, we should ask the question: Do we need both of them or can we pick one and call it the standard? @landitus need you here

PR #616, opened 2018-06-15
[WIP] breadcrumb component

beneliflo:
Improving breadcrumb component with route

PR #672, opened 2018-07-06
Improve breadcrumb component rebase

beneliflo:
Created new branch after last changes on master.

Tracking https://github.com/auth0/cosmos/pull/616


src/components/atoms/breadcrumb/breadcrumb.js
```diff
@@ -1,35 +1,115 @@
 import React from 'react'
 import styled from 'styled-components'
 import PropTypes from 'prop-types'
+import Link, { StyledLink } from '../link'
+import Icon from '../icon'
 
-import { spacing } from '@auth0/cosmos-tokens'
+import { withRouter, Route } from 'react-router-dom'
```
francocorreasosa: Question, do we require the user to be using `react-router-dom` to use the Breadcrumb component? @beneliflo @nkohari

nkohari: No; `react-router` shouldn't be a dependency of the main Cosmos library. It is a dependency of the manage example app, but that's it.

siddharthkp: Oooh! Good catch!

nkohari: Sorry, I missed this reply! Right now this is making the assumption that the crumbs exactly match the path segments, which I'm not sure is going to work.

Could we just make it visual, and handle clicks via `onClick()`, eg:

```jsx
<Breadcrumbs>
  <Breadcrumb onClick={...}>Home</Breadcrumb>
  <Breadcrumb onClick={...}>Parent</Breadcrumb>
  <Breadcrumb onClick={...}>Child</Breadcrumb>
  ...
</Breadcrumbs>
```

If we want to implement a special-case for different routing libraries, we could do that, but I don't think it can be part of the Cosmos core.

nkohari:
Closed in favor of #695.

PR #691, opened 2018-07-16
Default <input> to type=text

### LINKED ISSUE #681: TextInput: HTML type attribute should default to "text"
Right now, `TextInput` doesn't write a `type` attribute to the resulting `<input>`. We probably don't want to expose this as a prop, but for accessibility reasons we should output `type="text"` in the resulting HTML.

(via @jcenturion)

nkohari:
Closes #681.

### DIFF
--- a/core/components/atoms/text-input/text-input.js
+++ b/core/components/atoms/text-input/text-input.js
@@ -5,13 +5,13 @@ import { misc } from '@auth0/cosmos-tokens'
 import { StyledInput } from '../_styled-input'
 import { deprecate } from '../../_helpers/custom-validations'
 
-const TextInput = ({ defaultValue, ...props }) => {
+const TextInput = ({ defaultValue, type, ...props }) => {
   if (props.masked) {
     const length = defaultValue ? defaultValue.length : 8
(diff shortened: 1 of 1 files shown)


core/components/atoms/text-input/text-input.js
```diff
@@ -9,9 +9,9 @@ const TextInput = ({ defaultValue, ...props }) => {
   if (props.masked) {
     const length = defaultValue ? defaultValue.length : 8
     const maskedValue = new Array(length).join('•')
-    return <TextInput.Element {...props} placeholder={maskedValue} readOnly />
+    return <TextInput.Element type="text" {...props} placeholder={maskedValue} readOnly />
```
siddharthkp: This works perfectly 💯 

`<TextInput type="number"/>` renders a number input

Buuuut, do you think this might confuse folks, it looks like it's hard coded?

Suggestion: Add `type` in propTypes with `'text'` in `defaultProps`

nkohari:
Totally agree! I'll make the change.

nkohari:
I amended/force-pushed to keep things simple. WDYT now @siddharthkp ?

PR #695, opened 2018-07-17
Add multiple links to breadcrumbs

siddharthkp:
Supersedes https://github.com/auth0/cosmos/pull/672

Based on the [discussion in #672](https://github.com/auth0/cosmos/pull/672/files/5a4de096a1db52f39461163dcc22613374485f06#diff-65309c65c82b131312368482cd6d6761), I've made the API more declarative:

```jsx
<Breadcrumb>
  <Breadcrumb.Link href="/home">Home</Breadcrumb.Link>
  <Breadcrumb.Link href="/parent">Parent</Breadcrumb.Link>
  <Breadcrumb.Link href="/parent/child">Child</Breadcrumb.Link>
</Breadcrumb>
```

`Breadcrumb.Link` is a proxy to `Link` which means we get `href` and `onClick` both. (We also get the prop validations along with it)

`Breadcrumb` component adds the separators between them making it a purely visual layer.

I'm not sure what the left arrow in the component means, is it attached the first link or is the expectation that it takes you back one step?


landitus:
> I'm not sure what the left arrow in the component means, is it attached the first link or is the expectation that it takes you back one step?

Yes, the idea is to be able to have a "Back to" link available. We briefly discussed whether this should be part of the breadcrumb or not. It's a back link, but it's very close to the breadcrumb.

screen shot 2018-07-17 at 10 09 34

nkohari:
I think we might want to just let the user decide what icons to use with an `icon` prop. That way they can make the first crumb "home" if they want, and if they want to just have a single crumb with a back button (as in the third example) they can do so. (Although that could also just be implemented using a regular `Link`, to be honest.)

siddharthkp:
I like the Icon idea but does that also mean they can use the icon on each crumb? That sounds a little weird 🤔

siddharthkp:
Made all the changes

Check out chromatic preview or docs for preview

siddharthkp:
Made all the changes

Check out chromatic preview or docs for preview

siddharthkp:
screen shot 2018-07-18 at 8 11 30 pm

@landitus sup?

landitus:
I'm sending you over the correct icons for the breadcrumb

siddharthkp:
Cool, will wait

landitus:
You check the orgininal PR for the icons and the styles as well: https://github.com/auth0/cosmos/pull/672/commits/1330a94dc13739805825ae29d12fd113425c7aeb

siddharthkp:
Took the icons from https://github.com/auth0/cosmos/pull/672/commits/1330a94dc13739805825ae29d12fd113425c7aeb, had to add `top: -2px` to make it look good

landitus:
@siddharthkp Can you change the arrow icon for the home icon in this example https://www.chromaticqa.com/snapshot?appId=5aba7f6a07f95800202658ff&id=5b50d4e28221d100241ddf70 ?

This way, we are closer to merge.

landitus:
Tried to fix it in a supplemental commit. Also, fixed applying the same spacing to the separator and the icon. wdty of the code changes?

landitus:
FYI, I was trying to test this component in the `manage.dev` but got these kinds of errors (which we have to fix, for sure) but in master the errors are silent.

screen shot 2018-07-24 at 00 39 29

siddharthkp:
The error messages come from https://github.com/auth0/cosmos/pull/689, they are intentional so that our examples are always valid :)

You can temporarily close them with the x while working

landitus:
Did some styles polishing but couldn't get the link behavior to work. I added the `href` but it doesn't seem to work.

siddharthkp:
@landitus good catch, fixed it

Issue #682, opened 2018-07-09
Support forms inside of dialogs


nkohari:
Right now it isn't possible to create a `Dialog` with a `Form` inside it. This is a pretty important use case, so we should fix this problem soon. 😄 

The trouble is that `Dialog` has its own footer and buttons which aren't the same as the `Form` buttons. We want to support the use case of a `Dialog` that *doesn't* have a form embedded in it -- for example, we'll want OK/Cancel and Yes/No/Cancel dialogs at a minimum.

Worst case, we could introduce a new `FormDialog`, but there might be a better way to handle this problem.

landitus:
This was addressed in https://github.com/auth0/cosmos/pull/986

PR #700, opened 2018-07-19
Form inside Dialog

siddharthkp:
Fixes https://github.com/auth0/cosmos/issues/682

Not sure if I'm going in the right direction, could use some advice @nkohari @francocorreasosa 

Here's what a `Form` inside a `Dialog` looks like:

screen shot 2018-07-19 at 4 33 19 pm

Changes made: I've pulled out `Body` and `Footer` to make the `Dialog` more composable:

```jsx
<Dialog open={this.state.open} title="Example Dialog" onClose={() => this.setDialogState(false)}>

  <Dialog.Body>Are you sure?</Dialog.Body>

  <Dialog.Footer>
    <Button appearance="primary" onClick={ () => {} }>OK</Button>
    <Button appearance="secondary" onClick={ () => {} }>Cancel</Button>
  </Dialog.Footer>

</Dialog>
```

- No changes in chromatic previews 👌

- This also makes passing `Button`s the default way of composing actions, deprecating the `actions` prop.

- To make the form fit nicely inside, I passed the `width` prop with value `750`. We should probably have a way of converting this into a prop?

@landitus Should we have 2 sizes for the dialog? `default` and `wide`? `<Dialog size="wide">`?

_I thought about building `FormDialog` which has locks in `width={750}` and `actions=[]` inside it, but the wide form will be useful for other content as well like tables, videos, etc._

TODO:
- [ ] Update manage PoC
- [ ] Properly deprecate `actions` and `width` props

Bigger question here:

It feels we are converting every strictly built component with a `shape` type prop to a component that's more composable by exposing it's children.

Every time we do that, we are letting go of some _design decisions baked in_. This will mean people can use it in ways we don't want them to.

Should we be worried about this?

### DIFF
--- a/core/components/molecules/dialog/dialog-action.js
+++ b/core/components/molecules/dialog/dialog-action.js
@@ -1,9 +1,0 @@
-class DialogAction {
-  constructor(label, handler, appearance = 'default') {
-    this.label = label
-    this.handler = handler
-    this.appearance = appearance
-  }
-}
-
-export default DialogAction
--- a/core/components/molecules/dialog/dialog.js
+++ b/core/components/molecules/dialog/dialog.js
@@ -1,36 +1,12 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
-import Button from '../../atoms/button'
-import ButtonGroup from '../../molecules/button-group'
+
 import Overlay from '../../atoms/_overlay'
 import Icon from '../../atoms/icon'
 import Link from '../../atoms/link'
-import DialogAction from './dialog-action'
+import DialogFooter from './footer'
 import { colors, fonts, spacing } from '@auth0/cosmos-tokens'
-
-const createButtonForAction = (action, index) => {
-  // As we also support passing raw <Button> components
-  // as actions, we only need to create buttons for actions
-  // when the action is instance of DialogAction.
-  if (!(action instanceof DialogAction)) {
-    if (action.displayName !== Button.displayName) {
-      throw new Error('Invalid action component passed to Dialog.')
-    }
-
-    return action
-  }
-
-  const buttonProps = {
-    onClick: action.handler,
-    appearance: action.appearance
-  }
-  return (
-    <Button key={index} {...buttonProps}>
-      {action.label}
-    </Button>
-  )
-}
 
 const Dialog = props => (
   <Overlay {...props}>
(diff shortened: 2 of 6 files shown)


siddharthkp:
Added `size` prop, this will deprecate `width` prop

siddharthkp:
Made some tiny improvements to the form, good to go

There's some extra feedback from @landitus but that doesn't have to go in this PR

nkohari:
This looks good to me. I'm partial to opening the API (see #718), but you're right that it will mean a loss of control.

Is this okay to merge? I see some stuff was denied in Chromatic. Are those the changes you were talking about that can be handled elsewhere?

siddharthkp:
> Are those the changes you were talking about that can be handled elsewhere?

Yep, I think so 

@landitus Can you confirm this is good to go

nkohari:
@landitus Can you add an explanation of what needs to be done to get this merged?

landitus:
Sorry for the delay @siddharthkp !

I don't see `Form.actions` working inside the modal footer. The `Form.actions` have a lot of logic and styles tied to the form layout (with special alignment and margins). Inside the dialog, these buttons will display as in all dialogs (centered in the dialog window). I think we need to go on a different direction. I like making these more composable by using the buttons:

```
<Dialog.Footer>
    <Button appearance="primary" onClick={ () => {} }>OK</Button>
    <Button appearance="secondary" onClick={ () => {} }>Cancel</Button>
</Dialog.Footer>
```
To avoid adding the button spacing in the `Dialog.Footer`, we could wrap the buttons in a `ButtonGroup`

> To make the form fit nicely inside, I passed the width prop with value 750. We should probably have a way of converting this into a prop?

It's a pain to use the labels on the left inside the modal. Nowing we support both layouts, we need to find a way to make this form layout more responsive/elastic.  

> @landitus Should we have 2 sizes for the dialog? default and wide? <Dialog size="wide">?

Not use, a dialog window will probably need to be more flexible. I'm more inclined to do some named variations: default (), wide () and also be able to pass a custom number as the content can be very different.

All in all it seems we need to rethink what to do with the dialog. I believe we should have some design mockups first to check the variations.

siddharthkp:
> All in all it seems we need to rethink what to do with the dialog. I believe we should have some design mockups first to check the variations.

Is this PR n+1? If yes, let's merge it and keep improving it
If not, let's close it and start with the mockups

siddharthkp:
> The Form.actions have a lot of logic and styles tied to the form layout (with special alignment and margins). Inside the dialog, these buttons will display as in all dialogs (centered in the dialog window). I think we need to go on a different direction. I like making these more composable by using the buttons

I like the idea of using `Form.Actions` because it means users can keep using the same API. The centering/design should be invisible to the user without requiring extra effort

siddharthkp:
Closing this PR

It will be back 😄

PR #716, opened 2018-07-23
Added conditionals to styles for form layouts to keep them from leaking to each other

### LINKED ISSUE #715: Form layout label-on-top has wrong styles
When the form layout is set to `label-on-top` it still has some unnecessary styles left over from the `label-on-left`, like the min-height of the label and some margins and paddings.

screen shot 2018-07-23 at 10 40 07

landitus:
Fixes #715


landitus:
Can you take a look @siddharthkp ?

siddharthkp:
@landitus I see there's a rejected chromatic build, what's up?

landitus:
Oh, it’s not related to this PR per se. Just spotted something fishy with the dialog with a text and form example. The paragraph has weird styling and the link is in different blue and underlined. 

 Will add it as a separe issue so we can move forward

Issue #665, opened 2018-07-05
Compound components: Pass on props to inner elements 


siddharthkp:
There are some cases where we do not pass all the props through the DOM element

> I was not able to disable form elements and I couldn’t find how to do it on the docs
> – @tehsis

PR #718, opened 2018-07-24
Form actions pass props through

siddharthkp:
Fixes https://github.com/auth0/cosmos/issues/665

Added additional `props` prop that is forwarded to the `Button` component

```jsx  
<Form.Actions
  primaryAction={{ label: 'Save Changes', props: { disabled: true }, handler: () => {} }}
/>
```

Does this API look a little wonky? Should we use a more explicit name like `extraProps` or `buttonProps`?

&nbsp;

Demo: [form-actions#passing-button-props](https://auth0-cosmos-form-actions-pass-props-through.now.sh/#/component/form-actions#passing-button-props)

#### misuse 

There is little scope for misuse here, the `appearance` property comes after the spread props which means it will override it even if someone tries to pass `appearance="primary"` in a destructive action

```jsx
<Button {...destructiveAction.props} appearance="destructive" icon={destructiveAction.icon}>
  {destructiveAction.label}
</Button>
```

There is however room for indirection. These 2 blocks have the same effect:

```jsx
primaryAction={{ label: 'Save Changes', icon: 'home', props: { disabled: true }, handler: () => {} }}

primaryAction={{ label: 'Save Changes', props: { icon: 'home', disabled: true }, handler: () => {} }}
```

We pass the `icon` prop transparently to `Button` right now, but could be troublesome in the future

#### alternate syntax

We can offer the full control which is open to misuse but it supports all props directly because it is more composable

```jsx  
<Form.Actions>
  <Button appearance="primary" disabled={true} >Save Changes</Button>
  <Button appearance="primary">Two save buttons</Button>
</Form.Actions>
```

### DIFF
--- a/core/components/molecules/form/actions/actions.js
+++ b/core/components/molecules/form/actions/actions.js
@@ -30,10 +30,10 @@ const Actions = props => {
           <ButtonGroup>
             {primaryAction && (
               <Button
+                {...primaryAction.props}
                 appearance="primary"
                 icon={primaryAction.icon}
                 onClick={primaryAction.handler}
-                {...primaryAction.props}
               >
                 {primaryAction.label}
               </Button>
@@ -42,6 +42,7 @@ const Actions = props => {
             {secondaryActions &&
               secondaryActions.map((action, index) => (
                 <Button
+                  {...action.props}
                   appearance="secondary"
                   icon={action.icon}
                   key={index}
@@ -54,6 +55,7 @@ const Actions = props => {
             {destructiveAction && (
               <Right>
                 <Button
+                  {...destructiveAction.props}
                   appearance="destructive"
                   icon={destructiveAction.icon}
                   onClick={destructiveAction.handler}
--- a/core/components/molecules/form/actions/actions.js
+++ b/core/components/molecules/form/actions/actions.js
@@ -30,7 +30,7 @@ const Actions = props => {
           <ButtonGroup>
             {primaryAction && (
               <Button
-                {...primaryAction.props}
+                {...primaryAction}
                 appearance="primary"
                 icon={primaryAction.icon}
                 onClick={primaryAction.handler}
@@ -42,7 +42,7 @@ const Actions = props => {
             {secondaryActions &&
               secondaryActions.map((action, index) => (
                 <Button
-                  {...action.props}
+                  {...action}
                   appearance="secondary"
                   icon={action.icon}
                   key={index}
@@ -55,7 +55,7 @@ const Actions = props => {
             {destructiveAction && (
               <Right>
                 <Button
-                  {...destructiveAction.props}
+                  {...destructiveAction}
                   appearance="destructive"
                   icon={destructiveAction.icon}
                   onClick={destructiveAction.handler}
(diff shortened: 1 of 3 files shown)


nkohari:
Another open-ended/closed-ended question. 😄 

I don't really like the first option, using `props`. IMO, if we want to forward props directly, I think we should just forward the contents of the action object itself on to `Button`.

I'm sort of partial to the last one, but it does mean that form actions will no longer be an abstract concept. That is, they'll always be rendered as `Buttons`. That's probably not a problem, but I don't know if we can apply this same approach to all of the other places that actions are used. (For example, we've considered changing the `ResourceListItem` actions into a drop-down.)

landitus:
> I don't really like the first option, using props. IMO, if we want to forward props directly, I think we should just forward the contents of the action object itself on to Button.

I agree ! I think it would be best to pass all props directly to the `primaryAction` like:

```
<Form>
...
<Form.Actions
  primaryAction={{ label: 'Save Changes', disabled: true, type: 'submit', loading: false, onClick: () => {} }}
 secondaryAction={{ label: 'Cancel', disabled: true, type: 'button', loading: false, onClick: () => {} }}
/>
</Form>
```

We could filter setting the `appearance` prop, but ultimately you should be able to pass all properties for the button (or a subset). I'm not so much concerned about a misuse, as we are providing also documented guidelines on how this actions should be presented. It's more about saving time baking in decisions like spacing, button appearance while getting more control over the presentation.

One step further would be to create a template for this button shape so that it's easy to set and maintain. I'm looking at Polaris for this, and looks good: 

- https://polaris.shopify.com/components/structure/page#navigation
- https://polaris.shopify.com/components/structure/page-actions#navigation

In the end, it's about the Open/close debate, which I see valid points for both. We are using the closed component method selectively on the compound components to get more control over the styling, while always supplying the internal parts if you wish to extend it or [DIY.]

Just to make a fair comparison with the first option, the composable one could be:
```
<Form>
...
<ButtonGroup>
  <Button appearance='primary' disabled='true' loading='false' onClick=''>Save</Button>
  <Button disabled='true' loading='false' onClick=''>Cancel</Button>
</ButtonGroup> 
</Form> 
```
wdyt?

siddharthkp:
> I don't really like the first option, using props. IMO, if we want to forward props directly, I think we should just forward the contents of the action object itself on to Button.

Beautiful

```jsx
primaryAction={{ label: 'Save Changes', loading: true, handler: () => {} }}
```

This is now supported

core/components/molecules/form/actions/actions.js
```diff
@@ -30,6 +30,7 @@ const Actions = props => {
           <ButtonGroup>
             {primaryAction && (
               <Button
+                {...primaryAction}
                 appearance="primary"
                 icon={primaryAction.icon}
                 onClick={primaryAction.handler}
```
nkohari: I know we recently changed this already, but if we're just going to proxy props, I wonder if it makes sense to just change `handler` to `onClick`?

siddharthkp: Makes sense

Issue #692, opened 2018-07-16
Button has a bug when setting the URL prop


landitus:
The Button has an undesired visual change when the URL prop is set. Check the demo:

![kapture 2018-07-16 at 11 26 59](https://user-images.githubusercontent.com/239215/42764216-73131d9e-88eb-11e8-8cc3-3332be80f43a.gif)

siddharthkp:
Fixed in https://github.com/auth0/cosmos/pull/728

PR #728, opened 2018-07-25
fix bug with link appearance

beneliflo:
https://github.com/auth0/cosmos/issues/692


nkohari:
@landitus Can you review this so we can get it merged?

Issue #679, opened 2018-07-09
Spinner: Support different sizes and colors


nkohari:
We should support different sizes and colors for the spinner.

(via @jcenturion)

@landitus Can you add a mockup to this issue?

landitus:
**Mockup:**
<img width="744" alt="screen shot 2018-08-16 at 16 14 43" src="https://user-images.githubusercontent.com/239215/44229804-9222dc00-a16f-11e8-9cf8-1a4eb71ab46d.png">

**Sketch file:** 
https://www.dropbox.com/s/enlo8hp1iree33s/Spinner.sketch?dl=0

The `default` size is the larger one, the small one is size `small`

landitus:
Closing as this is all supported.

PR #730, opened 2018-07-25
added spinner size

beneliflo:
https://github.com/auth0/cosmos/issues/679

Added spinner size prop. Inspired by https://styleguide.auth0.com/#/components/spinner/stage
Maybe we should add auth0 avatar inside when the spinner is `medium` or `large` like in styleguide. WDYT?


nkohari:
@landitus Can you review this so we can get it merged?

PR #731, opened 2018-07-26
Global resets v2

siddharthkp:
- [x] always add custom line-height to cosmos components (`sc-`)
- [x] split resets into multiple parts
- [x] create `<ResetsWrapper>` component which applies component resets to any subset of the app

By default, all the resets are applied, so our chromatic snapshots should stay unchanged.

&nbsp;

I've split the resets into multiple parts, `ResetWrapper` applies the `componentResets` on demand.

### usage

```jsx
import {ResetsWrapper} from '@auth0/cosmos'

<App>
	<Sidebar>this is not effected</Sidebar>
	<ResetsWrapper>
		<Form>componentResets are applied here</Form>
	</ResetsWrapper>
</App>
```

&nbsp; 

@landitus  Can you check the resets and confirm these are the right rules: https://github.com/auth0/cosmos/blob/7753274721d0dfbf613e98878e132280c963a4bf/core/components/_helpers/globals.js

Should we change our manage PoC to manage app? (default mode off, in a different PR)


nkohari:
In order to get the right styles, is it required to wrap your components with a `<ResetsWrapper>` block? If someone doesn't do that, what happens?

nkohari:
Created #732 to track this in our project.

landitus:
We need to document the strategies to be able to merge this, so we are clear that we are addressing our customers scenarios. Like 1) App has global resets and uses Styleguide 2) App has no global resets (fresh new app)

Probably this needs an update too:

screen shot 2018-07-26 at 13 45 52

siddharthkp:
> In order to get the right styles, is it required to wrap your components with a <ResetsWrapper> block? If someone doesn't do that, what happens?

Cosmos components don't look like they should 😢 

> We need to document the strategies to be able to merge this, so we are clear that we are addressing our customers scenarios. Like 1) App has global resets and uses Styleguide 2) App has no global resets (fresh new app)

I guess the scenario worth mentioning is your app has custom resets (https://github.com/orgs/auth0/teams/design-system/discussions/24)

siddharthkp:
We don't need this right now. Closing to minimise clutter.

Will reopen if needed

PR #738, opened 2018-07-27
Add document playground for folks to report issues

siddharthkp:
Have a go at it: https://auth0-cosmos-docs-playground.now.sh/docs/#/playground


nkohari:
I like this idea. Should we add it to the sidebar navigation?

landitus:
Let's add it to the sidebar.

siddharthkp:
✅ Added to sidebar and in feedback section

PR #739, opened 2018-07-27
Add support for size prop to TextInput

### LINKED ISSUE #707: Add a size prop to TextInput which can support normal or small
This is primarily needed in the `PaginationToolbar`. We need to support a smaller version of `TextInput`.

* [x] Add mockup image (@landitus)

nkohari:
Closes #707.

This patch adds a `size` prop to `TextInput` which supports `default`, `large`, `small`, and `compressed` sizes that match those used in `Button`.

I also updated the `Stack` used in Chromatic tests in this patch, so it's possible there may be a few unexpected changes.

### DIFF
--- a/core/components/_helpers/story-stack.js
+++ b/core/components/_helpers/story-stack.js
@@ -1,25 +1,31 @@
 import React from 'react'
+import PropTypes from 'prop-types'
 import styled from 'styled-components'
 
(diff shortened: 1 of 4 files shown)


nkohari:
Created #740 to track issue with `Button` discovered here, but since it's unrelated to this work, I'm going to merge this as-is.

PR #747, opened 2018-07-30
Test scenario 2

siddharthkp:
This is to test scenario 2 from https://github.com/orgs/auth0/teams/design-system/discussions/24

It states that if you drop cosmos in an application that already has the legacy styleguide, it should work out of the box

The [chromatic tests](https://www.chromaticqa.com/build?appId=5aba7f6a07f95800202658ff&number=564) on this PR should pass to confirm this (they do 🎉)

### DIFF
--- a/examples/manage/public/index.html
+++ b/examples/manage/public/index.html
@@ -1,4 +1,9 @@
 <html>
+  <head>
+    <!-- Add legacy styleguide -->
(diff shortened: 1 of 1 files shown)


siddharthkp:
Closing this PR because we don't have to merge it, https://github.com/auth0/cosmos/pull/745 covers it

Issue #736, opened 2018-07-26
When clicking on the Table header to sort, the icon doesn't reflect the sort order 


landitus:

landitus:
Fixed in #758

PR #749, opened 2018-07-31
Improve table sorting

siddharthkp:
Added an example: [link](https://auth0-cosmos-add-sorting-example.now.sh/docs/#/component/table#sorting)

API changes:

- `Table` now accepts `sortOn` and `sortDirection`
- Removed `sort` in `Table.Column`

Visual changes:

1. Move `onClick` on header from `sortIndicator` to `TableHeader.Cell` (can't click on the sort indicator if it's not visible)

1. Add `sortable` check on `props.onSort` call (tiny bugfix)

1. Add hover state to sortable columns in header

Also fixes https://github.com/auth0/cosmos/issues/736

Bugs questions/issues:

- column width changes when sorting field changes 🐞 

- we should document `Table.Column` + add examples in the parent `Table`. `comparator`  is a really helpful but hidden feature 🙂

- should sorting work out of the box? For a table that does not have pagination + API requests, can we make the table sortable for the user without any extra effort? (And let the user take control by passing the `onSort` prop)


siddharthkp:
What do you think about the suggestion? Should I implement it?

francocorreasosa:
I really like that approach since it is closer to the behavior we want to have in the table. 
I'd also would like to hear what the rest of the team think about it.

core/components/atoms/icon/icons.json
```diff
@@ -68,18 +68,18 @@
...
         "M13.76 9.659a.524.524 0 0 1-.152.394c-.024.024-.06.017-.087.035L6.903 16.16a.53.53 0 0 1-.747-.749l6.273-5.754-6.274-5.755a.528.528 0 1 1 .748-.747l6.616 6.068c.028.02.065.013.09.038.11.109.156.254.15.398z"
       ],
       "width": 20,
       "height": 20
     },
-    "chevron-right-fill": {
-      "paths": ["M7.903 2L6 3.88 12.18 10 6 16.12 7.903 18 16 10z"],
-      "width": 20,
-      "height": 20
-    },
```
siddharthkp: :thinking: what's even happening here with `lint-staged`?

siddharthkp:
@francocorreasosa Made the API changes, also renamed some variables 👼

core/components/atoms/icon/icons.json
francocorreasosa: Weird 😅

nkohari: Ran into this myself. I tried a few different things to get the right stuff committed but gave up. 😞

landitus:
Just saw this visual bug on sorting Columns
kapture 2018-08-01 at 11 37 14

landitus:
> column width changes when sorting field changes 🐞
Yep, let's check on this, looks like a CSS bug

> we should document Table.Cell + add examples in the parent Table. comparator is a really helpful but hidden feature 🙂
Let's document (add example) this before merging

> should sorting work out of the box? For a table that does not have pagination + API requests, can we make the table sortable for the user without any extra effort? (And let the user take control by passing the onSort prop)
Let's be sure the table can be used without sorting, as many will need that. I'm not getting exactly what you propose, but it's related to dumb/smart components principles?

landitus:
Should this be labeled as breaking change? or sort if being maintained and deprecated?

francocorreasosa:
@landitus @siddharthkp Do you mean `Tab.Column`? Or is there a way to directly use `Table.Cell`?

landitus:
> API changes:
> 
> Table now accepts sortOn and sortDirection
> Removed sort in Table.Column

I was referring to `Table.Column`. 

Should this be labeled as breaking change? or sort if being maintained and deprecated?

francocorreasosa:
@siddharthkp This prop would be deprecated/removed now, right? 
https://github.com/auth0/cosmos/blob/2faa5d47d831ba4b46933750b6530f8962dcfb43/core/components/molecules/table/table-column.js#L10

siddharthkp:
Can we freely make `💔 breaking changes` to experimental components?

siddharthkp:
We don't need this anymore because https://github.com/auth0/cosmos/pull/758

PR #750, opened 2018-07-31
Pagination components improvements

francocorreasosa:
## Changes
- [x] Apply new design to `<Pager />`
- [x] Using `compressed` size for text inputs and buttons.
- [x] Check if handler function was not passed to `onPageChange` and fail silently
- [x] Tag components as `unstable`
- [x] Fix docs for all pagination components


siddharthkp:
Looks great!

There's one snapshot I don't know what to do with: https://www.chromaticqa.com/snapshot?appId=5aba7f6a07f95800202658ff&id=5b609b9097f2e500242da81f @landitus You want to take this one?

core/components/_helpers/pagination.js
```diff
@@ -37,7 +37,13 @@ export function changePageIfAppropiate(rawNextPage, total, perPage, handlerFn) {
   const pageCount = pagesFromItems(total, perPage)
   const nextPageExists = nextPage > 0 && nextPage <= pageCount
 
-  if (nextPageExists) handlerFn(nextPage)
+  if (nextPageExists) {
+    if (handlerFn) {
+      handlerFn(nextPage)
+    } else {
+      console.error('You must define an onPageChange function.')
```
siddharthkp: 🤔 why though?

Should we tackle this with propTypes instead?

core/components/molecules/pager/pager.js
```diff
@@ -1,47 +1,69 @@
...
+import Button from "../../atoms/button";
+import Icon from "../../atoms/icon";
+import { changePageIfAppropiate, totals } from "../../_helpers/pagination";
 
 const StyledPager = styled.div`
   display: flex;
   justify-content: space-between;
   align-items: center;
-`
+`;
```
siddharthkp: 😱 `;;;;;`

Some one needs to review their `precommit` hook :)

core/components/molecules/pager/pager.js
```diff
@@ -1,47 +1,69 @@
...
+
+const StyledButton = styled(Button)`
+  ${props =>
+    props.position === "left" ? "padding-right" : "padding-left"}: 11px;
+  ${props =>
+    props.position === "left" ? "padding-left" : "padding-right"}: 7px;
+
```
siddharthkp: I'm not super in the context, why do we need to modify the Button?

core/components/molecules/pagination/pagination.md
```diff
@@ -1,10 +1,11 @@
 ```meta
   category: Layout
   description: "Paginate resources with a handy pagination component"
+  unstable: true
```
siddharthkp: 👍

core/components/_helpers/pagination.js
francocorreasosa: This is more of a helper function check. Mostly to check if we don't pass the function from the prop to the helper function.

core/components/molecules/pager/pager.js
francocorreasosa: Ohhh. I don't know what happened there. Checking...

core/components/molecules/pager/pager.js
francocorreasosa: The icon does not play well with the Button as-is. So I had to adjust the padding.

francocorreasosa:
@siddharthkp Can give this a review again? Thank you!

PR #751, opened 2018-08-01
Don't wrap contents of Button

### LINKED ISSUE #740: Button text wraps to another line if compressed
As discovered in #739 after making tweaks to the `Stack` used in tests.

nkohari:
Closes #740.


landitus:
Some little details here, so I denied 3 tests:

The spinner is not vertically aligned with the text (it's off by very little). The icons looks perfectly aligned but not the spinner. 

screen shot 2018-08-01 at 12 21 25
screen shot 2018-08-01 at 12 20 29

Very picky but, can we improve the last test? It seems funky overflowing the container. I think it's meant to, just to show you can't have two lines of buttons? I'm not super sure, so wanted to point it out.

screen shot 2018-08-01 at 12 20 38

landitus:
Didn't realise the spinner alignment is not introduced in this fix. We can fix it quick or through another PR :)

PR #752, opened 2018-08-02
Proposal: Use standard prettier config to improve IDE plugin compatibility.

francocorreasosa:
This PR replaces the command line interface parameters to `prettier` in favor of a `.prettierrc` file that also allows IDE plugins like the one in Visual Studio Code to read our Prettier settings and format appropriately on save (if set).

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,3 @@
+singleQuote: true
+semi: false
+printWidth: 100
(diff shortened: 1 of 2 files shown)


nkohari:
I was considering doing this also yesterday! Could we just add it to the `prettier` key in `package.json` instead of having a separate file? It should work the same.

francocorreasosa:
No problem! I didn't considered that, doing it now.

landitus:
Such a nice improvement! 💥

PR #756, opened 2018-08-03
Prepare for 0.3.3

siddharthkp:
- [x] Add changelog for 0.3.1 - 0.3.3
- [x] Update version in root

There are 2 breaking changes: 

- **BREAKING**: Use `color` name instead of hex codes for `Icon` [#684]
- **DEPRECATED**: Replace input `error` prop with `hasError` [#690]


nkohari:
Let's merge #751 before releasing, since it's kind of an ugly problem. Just finishing up tests.

nkohari:
Also #757. 😄

Issue #704, opened 2018-07-19
Add sorting example to Table documentation


nkohari:
The `Table` component supports sorting by clicking the column headers, but we don't have an example in the documentation that shows how this functionality works. We should add one.

PR #758, opened 2018-08-06
Automatic sorting for Table

siddharthkp:
Built on top of https://github.com/auth0/cosmos/pull/749

This simplifies how the table API works, here are the 2 modes:

(check out the docs for a [nicer view](https://auth0-cosmos-automatic-sorting.now.sh/docs/#/component/table))

1. Automatic sorting
	You can add `sortable` prop to `Table.Column` that you want the `Table` to be sorted on and it will sorted itself + clicking on `sortable` columns on header will work = uses internal state

2. Controlled sorting
	Define a `onSort` function to call when a column header is clicked: **This should disable any default logic in the table, the expectation here is that the user will sort items and pass them back as props = user props (see example)** 

&nbsp;

Changed signature of `comparator` so that it matches the custom sort function syntax in javascript

```diff
- (column, row1, row2)` =>  {}
+ (row1, row2, column)` =>  {}
```

Also fixes https://github.com/auth0/cosmos/issues/704


core/components/molecules/table/table-header.js
```diff
@@ -67,6 +68,7 @@ TableHeader.Cell = styled.th`
 
 TableHeader.SortIndicator = styled.span`
   padding-left: ${spacing.xsmall};
+  visibility: ${props => (props.display ? 'initial' : 'hidden')};
```
siddharthkp: beautiful

PR #767, opened 2018-08-07
Adding stressed story for components

jcenturion:
### Badge

image

### Breadcrumb

image

### Button

image

### Label

image

### Select

image

### Tag

image

### EmptyState

image

### List

image

### PageHeader

image

### ResourceList

image

### Stack

image

### Table

image

### Tabs

image


nkohari:
I see that a Chromatic test was denied; what action needs to be taken to unblock this PR?

landitus:
Badges should we stressed with numbers. Doing it now!

Issue #683, opened 2018-07-09
AvatarBlock and EmptyState: Add support for a target if a link is specified


nkohari:
We currently support an `href` prop in `AvatarBlock` and a `helpUrl` in `EmptyState`, which will cause the component to display hyperlinks. However, there's no way to control the `target` attribute of the resulting `<a>` tag.

Either we should add a `target` prop to these components, or change the link prop to accept either a string or an object. That is, the link prop would support a type of something like:

```typescript
string | { url: string, target: string? }
```

PR #771, opened 2018-08-08
Support target in link for EmptyState and AvatarBlock

siddharthkp:
Fixes https://github.com/auth0/cosmos/issues/683

Supports both string and shape

```js
PropTypes.oneOfType([
  PropTypes.string,
  PropTypes.shape({
    href: PropTypes.string,
    target: PropTypes.string
  })
])
```

Concerns:

1. If we are accepting an object, will folks expect other props will be passed on as well? like `onClick` for link. Should we do that and pass the object forward as props? `<Link {...props.helpUrl} />`

2. The result in `AvatarBlock` is slightly weird because the prop itself called `href` but supports `href = {{ href: 'auth0.com', target: '_blank' }}`

    We probably need to standardise the prop name and call it `link` or something


nkohari:
Is this marked don't merge because it's still in progress, or are you just looking for reviews?

landitus:
The standard way then would be:
```
link = {{ href: 'auth0.com', target: '_blank' }}
```?

Which components would be affected to create a consistent API?

**Comment by siddharthkp** [comment]
@nkohari 

> Is this marked don't merge because it's still in progress, or are you just looking for reviews?

Sorry missed this

Not happy with the API, so definitely a work in progress. 

Also, looking for feedback to land on a good API


> I think that forwarding the entire object on to Link as props makes sense. I also agree that we should deprecate href on EmptyState and rename the prop to link for clarity.
> link = {{ href: 'auth0.com', target: '_blank' }}

Looks good to me!

**Comment by siddharthkp** [comment]
Looking good now: 

```jsx
<EmptyState
  icon="clients"
  title="Clients"
  link={{ href: 'https://auth0.com', target: '_blank' }}
/>
```

The object format also supports extra parameters which are ...proxied to the `Link` element 🎉 

Final question, do we want to support both string and object for `link`, Or should we lock down on object?

screen shot 2018-08-31 at 2 27 15 pm

landitus:
> Final question, do we want to support both string and object for link, Or should we lock down on object?

What's your take @siddharthkp ?

siddharthkp:
I'd say let's support both

Issue #732, opened 2018-07-26
Fix issues with global styles and resets in styleguide/css apps


nkohari:
We have noticed style overrides when using Cosmos in an application that currently has resets, the Styleguide and local globals styles  that affect globally buttons and inputs. 

We have talked about many strategies for the long term, but we want a hot and simple solution to apply now :) The first step is to test a solution in a branch in the product before merging and releasing a fix.

landitus:
The problem we noticed comes from differences in the line-height between Styleguide and Cosmos. This disparity can only grow, so it makes sense to explore how to introduce changes. So far, we have done some **in-product testing** with @cristiandouce and the **wrapper approach** seems to work: styles from the app remain untouched and Cosmos components get shielded and appear with the proper `line-height`.

### Test
1. Cosmos resets are disabled in the application.
2. A wrapper component with these are the exact CSS rules.

```
const CosmosResetWrapper = styled.div `
  /* Cosmos resets */
  * {
    box-sizing: border-box;
  }

  strong, em {
    font-weight: ${fonts.weight.bold};
  }

  body, input, textarea, button, select {
    font-family: ${fonts.family.text};
    font-weight: ${fonts.weight.normal};
    font-size: 14px;
    -webkit-font-smoothing: antialiased;
    -moz-osx-font-smoothing: grayscale;
  }

  /* Cosmos globals */
  [class^="sc-"] {
    line-height: ${misc.lineHeight};
  } 
`

const BackwardsCompatibilityStyles = styled(CosmosResetWrapper)`
  /* should we need to override stuff */
`

/* Usage 
<BackwardsCompatibilityStyles>
  ... <Pages>
</BackwardsCompatibilityStyles>
*/

```

 Let's try them in-product from an up-to-date branch before merging in.

### Test case
Any form with labels on the left.

<img width="597" alt="screen shot 2018-08-08 at 11 59 58" src="https://user-images.githubusercontent.com/239215/43845531-98ef5a36-9b02-11e8-85e1-ad926e0c041d.png">

siddharthkp:
Tacked in https://github.com/auth0/cosmos/pull/776

PR #776, opened 2018-08-09
Add resets wrapper

siddharthkp:
Fixes https://github.com/auth0/cosmos/issues/732

line


siddharthkp:
You mean breaking change now or later? Either way, this is limited to use by manage folks

siddharthkp:
The problem with current approach: 

How we use reset wrapper makes it ineffective, here's why:

Take this example code:

```jsx
<ResetWrapper>
  <Button>click me</Button>
</ResetWrapper>
```

this renders something like:

```html
<style>
  .reset .button { line-height: 1.6 } /* coming from wrapper */
  .button { line-height: 1.8 } /* coming from component */
</style>
<div class="sc-hash reset">
  <button class="sc-hash button"></button>
</div>
```

Because `.reset .button` is more specific than `.button`, it will get precedence. This is not how resets should work 🤔 

We can't inject them globally because we also need to scope them to the contents inside `ResetWrapper`

We have to go back to the original problem statement:

_When you drop cosmos components in an application with global variables_, they conflict with our globals. Example:

```css
/* global */
* { line-height: 2 }

/* cosmos-global */
sc-* { line-height: 1.6 }

```

If you keep `cosmos-globals`,
✅cosmos components are happy
❗️application's styled-components will get `line-height: 1.6` instead of 2

If you remove `cosmos-globals`,
❗️application's styled-components are happy
✅cosmos components get `line-height: 2` instead of 1.6

siddharthkp:
Alt solution in https://github.com/auth0/cosmos/pull/791

PR #782, opened 2018-08-09
Support free texts in Page Header, Empty State and Alert.

### LINKED ISSUE #753: Provide a way to add text with formatting in Empty State, Page Header, Alert.
Empty State, Page Header, Alert have a way to add text but no formatting is allowed. These components will need to allow text with either **bold**, [link](url) or sometimes `code`. Note that [Form field helper text](https://auth0-cosmos.now.sh/docs/#/component/form-field#help-text) has a way, but it's wonky. Maybe it can be reviewed as well.

screen shot 2018-08-02 at 16 45 49

Either way, it's interesting to check if some API decisions are shared or not.

francocorreasosa:
Resolves #753


core/components/_helpers/free-text.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+export function renderText(text, children, { useParagraph = false } = {}) {
+  if (children) {
+    if (useParagraph) return <Text>{children}</Text>
+    return children
+  }
+
+  if (typeof text === 'string') {
+    if (useParagraph) return <Text>{text}</Text>
+    console.log({ useParagraph })
```
nkohari: I know this is still WIP, but be sure to remove this. ;)

nkohari:
Could we handle this using an internal component called `Text`, which we'd just pass a spread of props to? For example:

```jsx
<Text {...props} />
```

The implementation of the component would be the same as the `renderText()` function, really, but the API might feel a little easier to work with. Just a thought, though, feel to disagree!

core/components/_helpers/free-text.js
francocorreasosa: Oops! Thanks

francocorreasosa:
Sounds good! 👍

francocorreasosa:
Done! API is much simpler now, thank you! @nkohari

siddharthkp:
Oops! Did we make a breaking change in this PR?

The old API no longer works!
```jsx
<PageHeader description={{
  text: 'Setup a mobile, web or IoT application to use Auth0 for Authentication.',
  learnMore: '/clients'
}}/>
```

Sorry for not catching it earlier!

francocorreasosa:
Hey @siddharthkp 👋 I tested it and it works.

image

siddharthkp:
🤔

I'm I confusing this with the children version?

Check out the .story.js just to be sure

francocorreasosa:
The children version was removed of this component. cc @landitus

siddharthkp:
Oh, should we _deprecate_ it instead of removing it? Just, so that we can say backward compatible?

francocorreasosa:
We did not reached a stage where the `children` version was public/released. So nobody got to use it and there is no reason to deprecate it

Issue #765, opened 2018-08-07
Change input border color to match Manage


siddharthkp:
Border color for inputs should be #CCC to match what we have in the product.

PR #786, opened 2018-08-10
Input color change in the tokens

landitus:
Closes: https://github.com/auth0/cosmos/issues/765

### DIFF
--- a/core/tokens/colors.js
+++ b/core/tokens/colors.js
@@ -34,7 +34,7 @@ const colors = {
   input: {
     background: '#FFF',
     backgroundReadOnly: '#F3F4F4',
-    border: '#D9D9D9',
+    border: '#CCCCCC',
     borderHover: '#B7B7B7',
     borderFocus: '#0a84ae',
     borderError: '#FF0000',


landitus:
This change of the border color in the inputs is not triggering any chromatic changes. Can you check @siddharthkp ?

siddharthkp:
Doens't look it's a bug in our code base!

Reproduced in https://github.com/auth0/cosmos/pull/812 and reported to chromatic folks

It seems like the color change is too subtle to be caught by visual tests /shrug

siddharthkp:
We can totally merge this PR though

landitus:
Ok, let’s

siddharthkp:
Got a reply. Yep it's a little too tiny for the diffing to catch

```
Hey Sid, I took a closer look at this and it seems like a false negative due to the anti-aliasing detection in our diff algorithm. I am going to be reworking this anyway in the next few weeks and I'll aim to solve this as part of that.

Cheers,
Tom
```

landitus:
Thanks for the update, and it’s good they are trying to solve it! No small change should pass, jajaja.

PR #790, opened 2018-08-13
Do not require form's primary action handler

francocorreasosa:
Question: In case a handler is still provided, should we automatically `preventDefault` so the form would not submit?


jcenturion:
@francocorreasosa I would leave the decision to `preventDefault` to the developer.

PR #791, opened 2018-08-14
[PoC] Base component

siddharthkp:
Fixes https://github.com/auth0/cosmos/issues/732

### Idea

Using a `BaseComponent` gives us the change to apply common styles to all our components.

In implementation, it attaches the same class to all components and drops the styles in it.

---

### Tests:

Failing test case with `ResetWrapper`: [playground link](http://bit.ly/2OzElMF)

Passing test case with `BaseComponent`: [playground link](http://bit.ly/2nEgCjg)

⚠️ [See tradeoffs with this approach](https://github.com/auth0/cosmos/pull/791#issuecomment-412792301)

Chromatic tests passed after making changes mentioned in tradeoff

Found a new bug from this: Tracking here https://github.com/auth0/cosmos/issues/792

---

### Syntax:

```diff
- import styled from 'styled-components'
+ import styled from '@auth0/cosmos/styled'

Button.Element = styled.button``
```

`@auth0/cosmos/styled` provides syntactical sugar that makes it easy to use for us (and people who will use cosmos/styled in their applications)

```js
import styled from '@auth0/cosmos/styled'

const Button.Element = styled.button``

// evaluates to

import BaseComponent from '../_helpers/base-component'

const Button.Element = BaseComponent.withComponent('button').extend``
```

 
---

### Context from #776 

The `ResetWrapper` solution doesn't work for us because of the extra specificity. 

```html
<style>
  /* wrapper: */
  .reset .button { font-weight: normal }

  /* component: */
  .button { font-weight: medium }
</style>

<div class="reset">
  <button class="button"></button>
</div>
```

Because `.reset .button` is more specific than `.button`, it will get precedence.


siddharthkp:
Here's the failing test case with `ResetWrapper`: [playground link](https://auth0-cosmos-reset-wrapper.now.sh/docs/#/playground?code=%250A/*%20you%20have%20access%20to%20all%20cosmos%20components%20here%20*/%250A%250Aconst%20Example%20=%20()%20=%3E%20%7B%250A%20%20return%20%3Cdiv%3E%250A%20%20%20%20%3CButton%3Efont-weight:%20medium%3C/Button%3E%250A%20%20%20%20%3CResetsWrapper%3E%3CButton%3Efont-weight:%20normal%3C/Button%3E%3C/ResetsWrapper%3E%250A%20%20%3C/div%3E%250A%7D%250A%250A%250Arender(Example)%250A)

siddharthkp:
Problems with this approach:

Nested styled-components elements don't inherit parent's style because they get reset.

```jsx
<Button.Element> // font-weight: ${fonts.weight.normal};
  <Text>Button</Text>
</Button.Element>
```

compiles to

```
<button class="base button">
  <span class="base text">Button</span>
</button>

and `.base` has `font-weight: ${fonts.weight.normal};`

siddharthkp:
Re: chromatic tests

There is a super unrelated change which adds `type=text` to `Input` 🤔, I have no clue where that is coming from

PR #819, opened 2018-08-16
Adding Card + CardList Component

jcenturion:
This PR introduces: 

- Card Atom
- CardList Molecule
- A little fix on `internal/docs/spec/prop-string.js` for supporting union type

### Card

```javascript
<Card
  disabled={false}
  checked={false}
  padding="small"
  background="gray"
  onClick={() => { }}
>
   Free content here
</Card>
```

### CardList

```javascript
<CardList
  columns={3}
  gutter="none"
  onItemClick={(e, item) => { }}
  items={[{
      content: 'Card'
    }, {
      content: 'Card'
    }, {
      content: 'Card'
  }]}
  renderItem={(item) => (
    <CardList.Item>
      Free content here
      {item.content}
    </CardList.Item>
  )}
/>
```

### Card Documentation

screencapture-localhost-8080-2018-08-16-10_41_18

### CardList Documentation

screencapture-localhost-8080-2018-08-16-10_42_10

### Card

card-sandbox

### CardList

cardlist-sandbox


andresgalante:
This PR introduces 2 componnents Card and Card List.

CardList should be a layout. I'd call it "GalleryLauyout" istead so it isn't usecase specific, a gallery can take any compoenent, not only cards.

The card itself needs a better structure, with a header, body and footer and better CSS, at the moment is not robust enough.

The Gallery layout should be a CSS grid contaier. Poisition items a layout with `inline-block` and margins is extremly fragile and wont support resposive states.

For accecibility, cards within a gallery should probably be `articles` and we might consider the gallery itself a list. There are probably other ARIA tags to consider in this view.

@jcenturion @landitus If you have questions I'd love to pair review this PR one and improve it before merging it.

francocorreasosa:
@landitus 	@siddharthkp 	@andresgalante 	What is the status of this PR? Should we merge it into an internal (up to date with master) branch and use it as a base to build this using the Grid component?

siddharthkp:
> @francocorreasosa please don't merge it. 

Agreed

> We'll provide a way to generate the same result in a flexible way with the cards component and the gallery layout.

I guess `GalleryLayout` will replace `CardLayout`. We can pick out the `Card` component and reimplement it if needed. (@landitus thoughts?)

andresgalante:
> I guess `GalleryLayout` will replace CardLayout. We can pick out the Card component and reimplement it if needed.

@siddharthkp the card component implementation in this PR isn't good. It needs heavily refactoring. It'll be much faster to just build it from scratch than trying to cherry-pick it from here and fix it.

As far as I know, we won't be working on new components till we fix what we have but if we need a card we can schedule it for next sprint, this is a simple component to add, shouldn't take more than a couple of days.

Also, I don't think we'll have a `CardLayout` as such, it'll be a composition of cards + gallery. We can probably show that as an example for users to follow.

Issue #800, opened 2018-08-14
Usability testing: Table


siddharthkp:
You can drop your feedback in this issue as comments and close it when you're done!

We'll create new issues for changes to be done.

Things to remember: How to pass data, column width + alignment, sorting, dynamic data/pagination

What makes a component _good_ for usage? See https://github.com/auth0/cosmos/issues/794

landitus:
I did a first pass at the table. I can flesh it out with examples or images.

- Bug: How do I set the initial sorted by column. I don’t know how to do it know. Document that example.
- Add support to easily set the alignment of each column's content (align: left, center, right)
- Add support for a loading state (with spinner). Check manage today for the example.
- Add support for an empty state (no data/items displayed)
- Each row has a pointer cursor, but it shouldn't.
- Let’s highlight each row on hover, but leave the cursor in default.
- Question: why can’t I add an element directly into the row (as a children)?
- Docs: Add example for table without any sorting
- Tests: Add test for table without any sorting

nkohari:
Converting these to issues. Some feedback:

1. From a UX standpoint, does it make sense to highlight a table row if clicking it doesn't do anything? Maybe we actually want to highlight the row if it has an `onClick`, or add a prop like `highlightOnHover`?
2. You can't (currently) add rows to `Table` as `children` because it's a fully-data-driven component. Letting users add rows manually would interfere with our ability to sort and page the table. We'd also have to re-do the API, because the children of `Table` are `Table.Columns`, rather than `Table.Rows` (and `Table.Cells`).

nkohari:
After looking back at the code, it looks like it was intended for the `cursor: pointer` style to be applied only if `onRowClick` was set, but there's currently a bug that's causing it to be set all the time. I'll fix the bug for now, but if we want to change how it works that's fine too.

nkohari:
I know that the original request was for table column alignment not to affect headers, but I think we may want to consider it. For example, the last two columns of this table are right-aligned:

![screen shot 2018-08-20 at 9 05 50 am](https://user-images.githubusercontent.com/1576/44342230-554d3280-a458-11e8-863a-3d0d66d9aea9.png)

landitus:
Thanks! Two things:
- Can we change and make the row highlight on hover, even if there is no onClick? This is the behaviour working today. I believe its a feature to make the row easy to read as you hover on it. The cursor would remain the default, as this PR makes sure.
[...]

PR #831, opened 2018-08-20
Add support for table column alignment

### LINKED ISSUE #825: Add alignment prop to table columns
There should be a prop to support alignment of column data to the left, center, or right. This should not affect the table header.

nkohari:
As requested in #800. Closes #825.

This patch adds an `align` property to `Table.Column` which allows the user to align the content of the columns left, center, or right. The original design was to keep the table headers aligned left, but after seeing the results, I made the `align` property affect the header as well.

### DIFF
--- a/core/components/molecules/table/table-column.js
+++ b/core/components/molecules/table/table-column.js
@@ -7,8 +7,13 @@ TableColumn.propTypes = {
   field: PropTypes.string.isRequired,
   title: PropTypes.string,
   width: PropTypes.string,
+  align: PropTypes.oneOf(['left', 'center', 'right']),
   sortable: PropTypes.bool,
   comparator: PropTypes.func
+}
+
+TableColumn.defaultProps = {
+  align: 'left'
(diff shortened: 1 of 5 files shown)


landitus:
Thanks! Two things:
- Can we change and make the row highlight on hover, even if there is no onClick? This is the behaviour working today. I believe its a feature to make the row easy to read as you hover on it. The cursor would remain the default, as this PR makes sure.
- The hover row color should be #FAFAFA. We can check in the tokens or add it to the Table with a comment to later tokenize it.

siddharthkp:
Should this also align the header in the same way? 

It looks slightly strange here:

preview

siddharthkp:
@landitus What do you think ^

landitus:
I'm still not sure how to solve the issue around the sorting arrows and alignment. I will take time to mockup a solution, maybe change the sorting arrow icons and behaviour, so we can take another shot at column alignment.

PR #850, opened 2018-08-22
Adding support for attributes for automation

jcenturion:
This PR introduces the `data-cosmos-key` attribute that will help with automation scripts.

### Tasks

- [x] Enable `data-cosmos-key` attribute to the components that will be used commonly in automation scripts
- [x] Add snapshot test to ensure that `data-cosmos-key` is being added
- [x] Add a new section in Cosmos docs to expose the Automation glossary

### Example

image

#### Sample queries

- `$('div[data-cosmos-key=breadcrumb]')` => selecting the breadcrumb container
- `$('div[data-cosmos-key=breadcrumb] a[data-cosmos-key=link]')` => selecting a breadcrumb link

### Implementation details

As you see the `data-attribute` is being added manually, this is because, IMHO, it does not make any sense to add something like [babel-plugin-react-add-property](https://github.com/alanbsmith/babel-plugin-react-add-property) to automatically add those attributes, since we don't need to add the attribute to every single component. Also, by adding the attribute manually we can control which components have it and if need add more in the future.

### Automation glossary

docs-automation

### DIFF
--- a/core/components/_helpers/automation-attribute.js
+++ b/core/components/_helpers/automation-attribute.js
@@ -1,0 +1,3 @@
+export default (defaultKey) => ({
+  'data-cosmos-key': defaultKey
+});
--- a/core/components/atoms/alert/alert.js
+++ b/core/components/atoms/alert/alert.js
@@ -6,6 +6,7 @@ import Link, { StyledLink } from '../link'
 import Paragraph, { StyledParagraph } from '../paragraph'
 import { Text } from '../../_helpers/free-text'
 import { deprecate } from '../../_helpers/custom-validations'
+import Automation from '../../_helpers/automation-attribute'
 
 const ReadMoreLink = styled(Link)`
   color: ${props => colors.alert[props.type].text};
@@ -48,7 +49,7 @@ class Alert extends React.Component {
   render() {
     if (this.state.visible) {
       return (
-        <Alert.Element type={this.props.type}>
+        <Alert.Element type={this.props.type} {...Automation('alert')}>
           <Paragraph>
             <em>{this.props.title}</em> <Text {...this.props} />
             {this.props.link && (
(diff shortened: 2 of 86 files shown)


siddharthkp:
> since we don't need to add the attribute to every single component. 

What's the downside of adding it to every component/element?


I'm a little scared of adding manual tasks because you need to remember while coding/code review (😢) or write tests to catch them when you forget (not so bad)

landitus:
Oh, we should add that template!

hzalaz:
You could easily add a test that iterates  every components and reports components with missing attributes. You could choose to make this fail the build (if you expect all components to have an attribute) or fail only in release, etc. 😄

jstrutz:
If we continue down this path, can we change the naming here to something a little more generic?  Perhaps `data-elem-ref`?

Reason is that we'll be using Cosmos to build larger, application-specific components, for which wouldn't make sense to also add a `data-cosmos-key` attrib because they're not in cosmos. We could use something else, but then functional test authors would have have to know which components were referenced by which attributes, adding friction to component refactors.

cristiandouce:
IMHO, same would happen by agreeing to something generic at this stage. Although we'd have more friction in the sense that'd you have to change it on every project using this "pseudo-standard".

By scoping the naming of the data key to the project you not only remove that hard responsibility but also could agree by semver some contract that if you want to change it you'd need a major release on the lib (which would make the project users and test builders aware).

The way I see it, making some automation testing in manage could look like this

`click('form[data-elem-key="manage.form"] button[data-cosmos-key="button"]')`

and I don't see much harm here.

cristiandouce:
Just adding a reference here: https://github.com/auth0/cosmos/issues/855

siddharthkp:
@jcenturion The snapshot tests are breaking on CI, they pass on local 🤔 

Seems like some components have extra styles on CI 😨

jcenturion:
@siddharthkp it seems to be that everything is green now.

PR #851, opened 2018-08-23
Add styled components plugin

siddharthkp:
> The goal is to make it easier for us to debug in development, since we can see which classes correspond to each component – Franco

Before:

```html
<div class="sc-fOZdva dicIel">
  <div class="sc-gVqHdG gLAkPy">
    <label class="sc-nVsqZ giXwJW">Name</label> 
  </div>
  <div class="sc-hwikVr elQzqq">
    <input class="sc-klEXDC ciDjpv" type="text"/>
  </div>
</div>
```

After: 

```html
<div class="field__StyledField-fOZdva dicIel">
  <div class="field__LabelLayout-gVqHdG gLAkPy">
    <label class="label__StyledLabel-nVsqZ giXwJW">Name</label> 
  </div>
  <div class="field__ContentLayout-hwikVr elQzqq">
    <input class="_styled-input__StyledInput-klEXDC ciDjpv" type="text"/>
  </div>
</div>
```


landitus:
@siddharthkp Can you elaborate and add information on what this PR will do and what's the usecase?

francocorreasosa:
@landitus I guess the goal is to make it easier for us to debug in development, since we can see which classes correspond to each component.

siddharthkp:
🤔 We should only apply this to dev, will make those changes

siddharthkp:
Err, chromatic seems stuck again :/

PR #859, opened 2018-08-24
Add table loading indicator

francocorreasosa:
## ✨ Added
- `loading` prop (boolean)


landitus:
Good point @siddharthkp ! I believe we can lay out some mockups for this to check what states should we support. You bring a good point about having an initial loading state and a different transition when loading pages from the pagination. Something similar to what @cristiandouce was concerned about.

siddharthkp:
@landitus @francocorreasosa Do we need to spec this out first?

siddharthkp:
- [x] Back merged master, fixed conflicts
- [x] Removed loading example in docs

Good to go as soon as the tests passed.

landitus:
Ok, will review now!

PR #865, opened 2018-08-27
Fix a few issues with uncontrolled mode table sorting, improve documentation

### LINKED ISSUE #829: Add support for a Table that isn't sortable
Right now `Table` will always render the sorting controls, even if none of the columns are marked as `sortable` and there is no `onSort` callback present. The default behavior for `Table` should be that it isn't sortable unless one of these props are set.

nkohari:
Closes #829 and fixes some extra issues covered in #800.


landitus:
@francocorreasosa wanna take a look and merge?

Issue #838, opened 2018-08-21
Form Input action is not vertically centered


francocorreasosa:
![image](https://user-images.githubusercontent.com/4152942/44409346-987ed280-a538-11e8-9cfe-9363dc9b0827.png)

PR #866, opened 2018-08-28
Fix action margins

francocorreasosa:
Fixes https://github.com/auth0/cosmos/issues/838

### DIFF
--- a/core/components/molecules/_action-input/_action-input.js
+++ b/core/components/molecules/_action-input/_action-input.js
@@ -6,18 +6,11 @@ import TextInput from '../../atoms/text-input'
 import Button from '../../atoms/button'
 import ButtonGroup, { StyledButtonGroup } from '../../molecules/button-group'
 import { multiply } from '../../_helpers/pixel-calc'
-import { spacing } from '@auth0/cosmos-tokens'
+import { spacing, misc } from '@auth0/cosmos-tokens'
 import { actionShape } from '@auth0/cosmos/_helpers/action-shape'
 
 /* TODO: width of button should be exported by button component */
 const widthOfButton = '36px'
-
-export const topMarginForActions = {
-  small: '0',
-  compressed: '2px',
-  large: '8px',
-  default: '6px'
-}
 
 const Wrapper = styled.div`
   position: relative;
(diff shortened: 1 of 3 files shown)


core/components/molecules/_action-input/_action-input.js
```diff
@@ -9,6 +9,14 @@ import { multiply } from '../../_helpers/pixel-calc'
 import { spacing } from '@auth0/cosmos-tokens'
 import { actionShape } from '@auth0/cosmos/_helpers/action-shape'
 
+const topMarginForActionsWithSize = size =>
+  ({
+    small: '0',
+    compressed: '2px',
+    large: '8px',
+    default: '6px'
+  }[size])
```
cristiandouce: Hey, minor question here: Is there any value on dynamically generating a new object on every function invocation?

landitus: Since this is calculated for the Form Field component, shouldn't this be stored near that component?

francocorreasosa: @cristiandouce Changing it now, there's no value on doing that. Thank you!
@landitus Yes, will move it to that one

core/components/molecules/_action-input/_action-input.js
```diff
@@ -12,6 +12,13 @@ import { actionShape } from '@auth0/cosmos/_helpers/action-shape'
 /* TODO: width of button should be exported by button component */
 const widthOfButton = '36px'
 
+export const topMarginForActions = {
```
cristiandouce: last one, I promise 😂 Should this be imported from `components/_helpers/action-spacing.js` instead of replicated?

francocorreasosa: Actually, I removed that file because we were only using it here 😛

landitus:
I've found something weird: the example in the docs looks different than the one in the test. The docs example still shows the icon off-centered. Can you check?

**Docs:**
https://auth0-cosmos-fix-form-action-centering.now.sh/docs/#/component/form-field
screen shot 2018-08-30 at 13 12 18

**Stories:**
screen shot 2018-08-30 at 13 13 34

PR #877, opened 2018-08-31
Fix problems with cosmos + styleguide setup

siddharthkp:
Just to see the visual diff

Found 2 bugs here: 
- We don't load semibold 
- Tabs has an invalid token

Tasks:
- Add styleguide + overview tests

Use https://github.com/auth0/cosmos/pull/879

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -5,3 +5,6 @@
   }
 }
 </style>
+<link rel="stylesheet" href="https://cdn.auth0.com/styleguide/core/2.0.5/core.min.css" />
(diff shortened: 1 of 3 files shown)


landitus:
The visual diff is correct. There are some differences when looking at Alerts for example, but more across the board. 

These issues are only visible when running the tests with the styleguide styles. So we are missing on them:
- https://github.com/auth0/cosmos/issues/895
- https://github.com/auth0/cosmos/issues/896

There probably are more. What would be the the next step here?

siddharthkp:
> What would be the the next step here?

Fix them one by one until chromatic is green again

siddharthkp:
Carried over in https://github.com/auth0/cosmos/pull/936

PR #907, opened 2018-09-10
Selector for cosmos components

siddharthkp:
Right now, we have attach `line-height: 1.6` to all components created with styled-components using a `sc-^` selector (the styleguide uses a different `line-height`)

This also spills over to the components in the application created with `styled-components` that are not from cosmos.

With this PR, we add a class to all cosmos components (`cs`) and use that to add `line-height: 1.6` to that

Even though this is a bug fix, it is also a breaking change because `line-height: 1.6` is no longer applied to the application's styled-components. If an application depended on this, this might (will?) change some styles

- [x] Add className "cs" to all cosmos components
- [x] Replace sc-^ with cs in globals

Links:

1. [Chromatic build for manage example out of the box](https://www.chromaticqa.com/snapshot?appId=5aba7f6a07f95800202658ff&id=5b95e15c645723002436ed86) ❗️
2. [Chromatic build for manage example with auth0/cosmos/styled instead of styled-components](https://www.chromaticqa.com/build?appId=5aba7f6a07f95800202658ff&number=839) ✅

### DIFF
--- a/core/codemods/replace-styled-components.js
+++ b/core/codemods/replace-styled-components.js
@@ -1,0 +1,15 @@
+const transformer = (file, api) => {
+  const j = api.jscodeshift
+
+  return j(file.source)
+    .find(j.ImportDeclaration)
+    .forEach(path => {
+      path.value.source.value = path.value.source.value.replace(
+        'styled-components',
+        '@auth0/cosmos/styled'
+      )
+    })
+    .toSource()
+}
+
+export default transformer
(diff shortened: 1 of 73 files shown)


siddharthkp:
@landitus @francocorreasosa Is there something we should change in this?

core/codemods/replace-styled-components.js
```diff
@@ -0,0 +1,15 @@
+const transformer = (file, api) => {
```
francocorreasosa: Codemods ❤️

siddharthkp:
Oh boy, this might be the golden bullet 

Killing this

Issue #857, opened 2018-08-23
Ability to truncate text content inside table cells to avoid multiple lines


landitus:
We should add support to truncate long text (if desired) and avoid having multiple lines. We could think if this is via a helper component or a prop to the column. This is the current behaviour in Manage.

<img width="878" alt="screen shot 2018-08-23 at 13 54 19" src="https://user-images.githubusercontent.com/239215/44540034-15967d00-a6dc-11e8-9ce7-3ca874b0eedd.png">

Also, let's stress test the cell content with and without truncation in Storybook to check how the table behaves with long content inside the cells. This could include the Avatar block, Labels, link and just plain text.

PR #914, opened 2018-09-11
Fix tests

francocorreasosa:
Fix tests for changes introduced prior to merging component tests to master.

PR #916, opened 2018-09-12
Truncate table cell text

### LINKED ISSUE #913: Add documentation for Table.Column
We should add docs and a Props Table for the `Table.Column` to be able to show the props and usage of the Column. This one, we can showcase the sorting properties and many more.

francocorreasosa:
Fixes https://github.com/auth0/cosmos/issues/857
Fixes #913 

image
~Need to merge #914 and rebase #915 from master to make tests pass here.~ Done


francocorreasosa:
Can you take a look? @landitus & @siddharthkp (I added the table to all `<table>` in the docs)

francocorreasosa:
Oops!  Forgot to update the `stressed with truncating` story. Now it's fixed, sorry. Please note that by default, truncating is not being applied, so the normal stressed story won't have it applied anyway.

landitus:
Would `truncate` be a better prop name than `truncating`?

francocorreasosa:
For me it's the same, I can change it if you want. We could use `truncated` as well.

landitus:
Let's go with `truncate` as it's shorter and more common.

francocorreasosa:
Done 😄

landitus:
I've noticed that when setting a truncated column combined with a percentage width, the table breaks.  It looks like it's not simple to make this work in tables. My guess is that it's easier to set percentages for widths, and also having two properties that break is not a good idea. 

From these answers, seems there are a bunch of options:
- https://stackoverflow.com/questions/5239758/css-truncate-table-cells-but-fit-as-much-as-possible

Seems many rely on hacks... which is not a good sign.

siddharthkp:
@francocorreasosa This has a bunch of unrelated chromatic changes 😢 

Can you replicate this PR from a fresh master? 🙏

francocorreasosa:
Yeah it’s weird. I think there are unapproved changes on master as well. Will review when I get to the office

francocorreasosa:
@siddharthkp I think in this kind of cases is better to _request changes_ than applying the `don't merge` label. Just for the sake of not mixing the purposes (I've seen `don't merge` is mostly used on WIP and _opened for staging` PRs.

PR #934, opened 2018-09-24
Allow to have nothing selected on Select

### LINKED ISSUE #925: Select does not let you have nothing selected
_via @astanciu_

**Describe the bug**
When using the Select component, you always need to have an item as the current selected one.

**Expected behavior**
By default nothing is selected.

francocorreasosa:
## Default
image

## With custom empty option text
image

Fixes #925

### DIFF
--- a/core/components/atoms/select/select.js
+++ b/core/components/atoms/select/select.js
@@ -21,7 +21,7 @@ const Select = ({ options, ...props }) => {
     <StyledSelect {...props} {...Automation('select')}>
       {/* First option will be selected if there is no value passed as a prop */}
       <option disabled hidden selected={!props.value} value="" {...Automation('select.option')}>
-        {props.emptyLabel}
+        {props.placeholder}
       </option>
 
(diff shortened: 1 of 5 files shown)


core/components/atoms/select/select.js
```diff
@@ -19,6 +19,9 @@ const Select = ({ options, ...props }) => {
 
   return (
     <StyledSelect {...props} {...Automation('select')}>
+      <option disabled selected={!props.value} value="">
```
siddharthkp: what does `!props.value` do 🤔

francocorreasosa: If we have a `props.value` set, this won't be the selected option, the `value` value will be

siddharthkp: 😅 got it

can you leave a comment there, looks like it has something to do with boolean values !false = true

francocorreasosa: Sure thing! Done.

PR #936, opened 2018-09-24
Add styleguide to storybook

siddharthkp:
Add styleguide + `COSMOS_DISABLE_RESETS` to storybook

@landitus This is interesting 🤔 

The chromatic tests are passing after adding styleguide to our storybook setup as long as we disable resets. No manual changes needed

I'm guessing there might be some untested things that look different/weird. Can be tested here: https://auth0-cosmos-cosmos-with-styleguide.now.sh

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -5,3 +5,6 @@
   }
 }
 </style>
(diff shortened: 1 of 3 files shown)


siddharthkp:
This PR looks suspicious, starting over in https://github.com/auth0/cosmos/pull/937

PR #937, opened 2018-09-24
Add local styles for styleguide compatibility

siddharthkp:
Test setup
- Add `styleguide/core/2.0.5/core.min.css` 
- Add `components/2.0.0/components.min.css`
- Add `FaktPro-Normal.woff2`
- Set `COSMOS_DISABLE_RESETS: true`

Fixes:
- Load `FaktPro-SemiBold.woff2` even with `COSMOS_DISABLE_RESETS`
- `p` will not get styled anymore, use `<Paragraph>`
- `strong` will not get styled anymore, use `<Text type="strong">`
- Rename `Text` from `free-text.js` to `FreeText`
- override default `ul` styles (list-style + padding-left + margin-bottom) for `DangerZone`
- Add `font-weight: normal` for label in `Checkbox`
- Use `box-sizing: content-box` for tick in `Checkbox`
- Add `display:block` for `Textarea`

Bugs found on master:
- Fix image extra padding
- line-height: 1.6 is applied to inline elements as well

&nbsp;


core/components/_helpers/globals.js
```diff
@@ -184,6 +184,17 @@ if (includeGlobals) {
       font-style: normal;
     }
 
+    @font-face {
+      font-family: fakt-web;
+      src: url(https://cdn.auth0.com/website/ds/fonts/fakt/FaktPro-SemiBold.woff2)
```
francocorreasosa: Should we use `url("...")` instead of `url(...)`?

.storybook/webpack.config.js
```diff
@@ -4,8 +4,8 @@ module.exports = storybookBaseConfig => {
   const plugins = storybookBaseConfig.plugins
 
   // plugins.push(
-  //   new webpack.DefinePlugin({
-  //     'process.env': { SKETCH: process.env.SKETCH }
+  //   new webpack.EnvironmentPlugin({
+  //     COSMOS_DISABLE_RESETS: true
```
francocorreasosa: Is this the way to switch when to include styleguide or not?

siddharthkp: Yeeep.

core/components/_helpers/globals.js
siddharthkp: Makes sense!

PR #939, opened 2018-09-25
New spinner variants

### LINKED ISSUE #594: Add large spinner (with logo) to Spinner component
The large spinner is primarily used for loading states between pages. The existing small spinner is used for loading states inside buttons.

I believe the large spinner should be the default one, so that would make a breaking change.

### Actual implementation
https://auth0-cosmos.now.sh/docs/spinner

### Design reference
screen shot 2018-06-08 at 16 24 55

### The API for the spinner could be:

**size: enum**
`default`: spinner with logo 
`small` : spinner without logo

For the light and dark version at first I thought of an `appearance` prop, that defaults to `light`? wdyt?

The docs could use some love and we will now have a Props table in there!

francocorreasosa:
Fixes #594
image
image

### DIFF
--- a/core/components/_helpers/default-prop-change.js
+++ b/core/components/_helpers/default-prop-change.js
@@ -1,0 +1,17 @@
+export default function defaultPropChangeWarning(
+	Component,
+	propName,
+	currentValue,
+	futureDefault,
+	version
+) {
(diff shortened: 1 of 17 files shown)


francocorreasosa:
Fixed backward compatibility

PR #947, opened 2018-09-26
Fix button vertical centering and introduce right icon alignment

### LINKED ISSUE #901: Add support for a right side icons to Buttons
When creating buttons for a pager component, there is the need to add a right > to the right. Button only supports an icon on the left by design. Should we consider adding an extra prop? or an icon alignment prop?

### LINKED ISSUE #902: Chevron icons are not centered in the icon bounding box
thanks for reporting @jcenturion 

image

francocorreasosa:
Fixes #901 
Fixes #902

> Thanks @andresgalante for suggestions in CSS 🙌


core/components/atoms/button/button.js
```diff
@@ -224,9 +227,11 @@ Button.Element = styled.button`
   pointer-events: ${props => (props.disabled || props.loading || props.success ? 'none' : null)};
   transition: border-color ${misc.animationDuration}, background ${misc.animationDuration};
 
-  ${Icon.Element}, ${StyledSpinner} {
-    position: relative;
-    top: -1px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
```
andresgalante: I don't think you need to justify this content.

To center the content in the btn, I'd change the padding. The current padding left and right is `0`. I'd make it one of the spacers of the system. 

I'd also remove the `min-width` or leave it ad safety net if you want.

core/components/atoms/button/button.js
```diff
@@ -224,9 +227,11 @@ Button.Element = styled.button`
   pointer-events: ${props => (props.disabled || props.loading || props.success ? 'none' : null)};
   transition: border-color ${misc.animationDuration}, background ${misc.animationDuration};
 
-  ${Icon.Element}, ${StyledSpinner} {
-    position: relative;
-    top: -1px;
+  display: flex;
```
siddharthkp: @francocorreasosa This changes the height of the buttons as well, we might have to tweak those

(chromatic shows the diff)

landitus:
> @landitus Out of scope from this PR, but while reading the code I was thinking that I'd also remove the line height set to 1. You never know when the content is going to wrap and having a line height is always important.

That's true, not going to help with responsiveness either! 

> There is also a strange display table on btn anchors to review. it's really odd.

> And one more, removing outlines for focus is really dangerous, especially if there is no other indication that the item has focus. It's like removing the wheelchair to a person that can't walk.

Thanks @andresgalante ! This is super helpful!!! I think the button now has a lot of layered code and it would be wise to review this and the rest of the comments. 👌

francocorreasosa:
Thank you again @andresgalante! I'm implementing your comments to check how the button looks like then. 😄

andresgalante:
To be honest, buttons should be `inline-block`. If the icons and the text share the same height and line height, then you wouldn't need to make the btn a flex container, things would align correctly anyway. Give me a few minutes, let me give that a try.

francocorreasosa:
I think the text and the icon won't always share height/line-height.

core/components/atoms/button/button.js
francocorreasosa: Yep, looking into that right now

andresgalante:
Ok, so... sorry for the confusion, now that I am actually changing the code I can see things clearly.

The button should not be a flex container, I am sorry for the misleading :(

Let's bring back:

```css
btn {
  display: inline-block;
  vertical-align: middle;
}
```

You may remove line 205, since box sizing is an inherited propierty and it's already set system wide.

Leave the selector for a margin right to for the right icon is correct.

Now, there is a larger issue, and I definitely don't want to be disruptive, especially not to fix this bug with the icons.

Once you merge this PR the chervron will align but it will misslaign the other icons.

The icons bounding box and the text box (meaning the size of the text times the line height) shuld be the same. It's important to teach designer how to create icons and deliver assets as well as to pay attention when they get implemented.

The line height value should never have a unit (for example not 48px). A line height is a ratio of the value of the font size. I can see that the default line height of the system is `1.6` and that's ok, there is no need to reset it for the btn. You could just delete that declaration since there is a `min-height`.

Ideally a button should not have a `min-height` but it should have paddings, and those paddings should be the same as in the form controls like inputs. Therefore everything will align correctly in the case of inline forms, or even inline forms with btns with icons.

To solve #901 the selector is adding the margin and it's perfect.

Now to solve #902, at this momement all icons are missaligned to the top.  If you remove the `top:-1px` then the'll missaligned to the bottom but the chevron will align correctly. If you'd like to hack it up and just align the chevrons, you could do something as:

```css
[name^="chevron-"] {
   top: 0
}```

But it's a horrible and temporary solution and it'll create technical debt.

I am sorry I can bring a definitive solution for this. Forms and buttons are some of the fundamental building blocks of the system, and if you don't nailed them, the lego bricks wont clock together. IMO this eserves to take a couple of days and clean up form controls together with btns.

and once again, sorry @francocorreasosa for generating more noise, specially when I opened pandora's box.

**Comment by francocorreasosa** [comment]
We have some cases for the `min-width`
image

File: core/components/atoms/button/button.js (hunk 1 above)
Comment (francocorreasosa) [reply core/components/atoms/button/button.js]: PD: We have some cases for the `min-width`

image

**Comment by francocorreasosa** [comment]
@siddharthkp @landitus can take a look again? thanks!

**Comment by landitus** [comment]
> Now to solve #902, at this momement all icons are missaligned to the top. If you remove the top:-1px then the'll missaligned to the bottom but the chevron will align correctly. 

I see the icons missaligned as well. So here are new chevron icons with the fixed alignment. This would mean we can remove the -1px position hack and all icons would look aligned inside the button, right?

[chevrons.zip](https://github.com/auth0/cosmos/files/2426247/chevrons.zip)

**Comment by andresgalante** [comment]
@lancelaird not really. due to the way the btn is been measured: the height of the btn is the `height` and `line-height`, and the lack of padding top of buttom plus the fact that the icon and text bounding boxes don't match, you end up with a sub-pixel issue that makes the button itself been misaligned.

The text is shift 2 pixels to the top.

The icon with the top -1 is not aligned with the text and if you remove it, it doesn't align either.

screen shot 2018-09-28 at 12 28 45 pm

There is no easy way out of this one. It's one of those cases where you have a scoop of 💩and the only way out is to eat it and refactor the way you measure height is forms and btns, as well making a base decision on line heights and icons.

**Comment by landitus** [comment]
I see. I would have to check the icon file first cause it's part of a library (buddicons). I've seen problems with sub-pixel positioning due to how the SVG is aligned inside the bounding box. This is at least one step to improve. 

The way the button is constructed to match a certain height (and multiple heights) is indeed the root of the problem. @andresgalante what do you suggest for a future refactor if we want to enforce a button/input height with items that will overflow the content's line-height?

**Comment by andresgalante** [comment]
> what do you suggest for a future refactor if we want to enforce a button/input height + having icons or spinners that will overflow the content's line-height?

I suggest to start from the base:

- Improve the global spacers system and define how you are going to use tokens consistently across the project.
- Make sure the icon bounding box is the same height as the text box. Meaning the size and viewBox of the svg should be the same as the `font-size` * `line-height`. Most often than not line height in buttons and titles is smaller than the ones in normal text. The system should have definitions for that at a global scale (As in line height global variables or tokens).
- Once that is set, then buttons, inputs, and selects can be modified: 
  - Remove all heights or min-heights declarations;
  - Remove all references to line height in pixels and use a local token that it's set to a global token (if there is a concept of scoped and global variables);
  - Add paddings, this values should be set in tokens to and consume global tokens;
  - All declarations of border, paddings, font size and line height should be the same in inputs, selects and buttons.

Then you'll have verical alignment within each component and among them, and you'll be happy ever after. 💖

**Comment by andresgalante** [comment]
Here is an example of what I've described:

https://github.com/andresgalante/cosmos/tree/andres-btns

There are some comments on the code and I turned on outlines for preview. You can clearly see how the chevron svg is the problem and not the btn itself.

Again take this as it is, I have no understanding of the system or the codebase.

**Comment by andresgalante** [comment]
@lancelaird after inspecting the chevron svg you are providing I can see where the problem is.

The vector is slightly off center and off pixel. The chevron shape is 13.31 x 7.76 px and it's not centered. Probably if you make it 14 x 8px with a 20px square artboard it'll help. Icons should follow a grid.

Here are some tips: https://medium.com/sketch-app-sources/exploring-ways-to-export-clean-svg-icons-with-sketch-the-correct-way-752e73ec4694

File: core/components/atoms/button/button.js [hunk 3]
```diff
@@ -170,16 +170,17 @@ const ButtonContent = props => {
...
 
+  if (props.text) {
+    content.push(<Button.Text key="text">{props.text}</Button.Text>)
+  }
+
   if (props.loading) {
-    content.push(
-      <Spinner key="spinner" size="small" inverse={getAttributes(props).loadingInverse} />
-    )
+    content.unshift(<Spinner key="spinner" inverse={getAttributes(props).loadingInverse} />)
```
siddharthkp: This logic is slightly difficult to read

Can we make all these conditionals independent?

```
if (props.text) ...

if (props.loading) ...

if (props.icon && !props.loading) ...
```

francocorreasosa: Yep 👍

landitus:
Seeing the Chromatic snapshots, there is some sort an optical alignment that’s not working here. Let me explain: Elements might be off-centered by looking at the numbers, but if you look at the `baseline` image, the icon is vertically aligned to the text, but looks misaligned in the `new` snapshots.

This could be because of the icon shape/artwork thing (mentioned above) or even the typography file and how it's rendered. Check the `spinner` inside the button as well, that it's not an icon, and the problem persists. So maybe it's not the icon shape after all.

**Looks aligned**
screen shot 2018-11-01 at 10 49 34

**Not so much**
screen shot 2018-11-01 at 10 49 55

PR #958, opened 2018-09-28
More prominent version switcher for docs

landitus:
Improved the switcher to be more prominent :)

screen shot 2018-09-28 at 19 47 13

### DIFF
--- a/internal/docs/docs-components/navigation.js
+++ b/internal/docs/docs-components/navigation.js
@@ -50,6 +50,17 @@ const Navigation = styled.nav`
   }
 `
 
+const LogoName = styled.h1`
+  font-size: 14px;
+  letter-spacing: 1.4px;
+  display: inline-block;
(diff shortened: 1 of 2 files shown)


francocorreasosa:
In Safari (macOS Mojave) looks like this:
image

In Firefox (Dev Edition):
image

Is this expected?

landitus:
Yep, we are using the native select, so I expect it to look different cross-browser. I believe that to achieve a unified look, we have to style the select with a custom select (hide the native, recreate the select behaviour). So it's something we can improve when we get to the custom select.

francocorreasosa:
Just noticed this also happens in our Select component. Maybe we can reset that with `-webkit-appearance: none;`? What do you think?

francocorreasosa:
Looks goooooood! 🙌

landitus:
Thanks!!

PR #959, opened 2018-10-01
Tooltip in button group UI fixes + Tooltip stories

### LINKED ISSUE #764: Unit tests for tooltip
Testing tooltips is tricky

Because it's only visible in hover, our visual tests don't catch it!

francocorreasosa:
Fixes #764


francocorreasosa:
@landitus I think this is ready to merge after you review it 👍

landitus:
Looks good to me! I would only add one more test, which is when the button with icon invokes the tooltip. Like: https://auth0-cosmos.now.sh/docs/#/component/button#adding-tooltips

This is a quick shortcut that only covers the top alignment, but we can test it too.

francocorreasosa:
Sure thing!

francocorreasosa:
Done. @landitus

Issue #942, opened 2018-09-25
Match Alert close button to styleguide


landitus:
Cosmos:
<img width="89" alt="screen shot 2018-09-25 at 13 30 27" src="https://user-images.githubusercontent.com/239215/46028702-ad1a3180-c0c7-11e8-845f-b54d13f80127.png">

Styleguide:
<img width="152" alt="screen shot 2018-09-25 at 13 30 22" src="https://user-images.githubusercontent.com/239215/46028707-b0152200-c0c7-11e8-855f-15bac23a0d2d.png">

PR #963, opened 2018-10-03
Match alert close button styling to styleguide - Issue #942

mrewers:
Addresses the issue: "Match Alert close button to styleguide #942" by setting the close icon to 20% opacity and black with 50% opacity on hover.

### DIFF
--- a/core/components/atoms/alert/alert.js
+++ b/core/components/atoms/alert/alert.js
@@ -122,7 +122,8 @@ Alert.Element = styled.div`
     opacity: .2;
     padding: ${spacing.small} ${spacing.small};
     &:hover {
-      color: rgba(0,0,0,0.50);
+      color: rgba(0,0,0,0);
+      opacity: 0;
     }
   }
 `
--- a/core/components/atoms/alert/alert.js
+++ b/core/components/atoms/alert/alert.js
@@ -119,7 +119,11 @@ Alert.Element = styled.div`
     right: 0;
     top: 0;
     color: ${props => colors.alert[props.type].text};
+    opacity: .2;
     padding: ${spacing.small} ${spacing.small};
+    &:hover {
+      color: rgba(0,0,0,0.50);
+    }
   }
 `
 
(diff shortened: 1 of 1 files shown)


core/components/atoms/alert/alert.js
```diff
@@ -119,7 +119,12 @@ Alert.Element = styled.div`
     right: 0;
     top: 0;
     color: ${props => colors.alert[props.type].text};
+    opacity: .2;
     padding: ${spacing.small} ${spacing.small};
+    &:hover {
+      color: rgba(0,0,0,0);
```
francocorreasosa: Hey @mrewers! Thanks for your contribution. Here you are setting the cross not to be visible on hover. That is not the behavor we would want for this state.

image

mrewers:
@francocorreasosa Apologies for the error, that's certainly not what I meant to do.

core/components/atoms/alert/alert.js
```diff
@@ -119,7 +119,12 @@ Alert.Element = styled.div`
     right: 0;
     top: 0;
     color: ${props => colors.alert[props.type].text};
+    opacity: .2;
     padding: ${spacing.small} ${spacing.small};
+    &:hover {
+      color: rgba(0,0,0);
```
andresgalante: Can't this just be `rgba(0, 0, 0, .5)` instead? and remove the `opacity` propierty?

@francocorreasosa shouldn't a color declaration be a token?

mrewers: Yeah you're right, I should probably have imported the black color token and used that instead. In which case it would be:
`
&:hover {
      color: black;
      opacity: .5;
`

I think the opacity line is needed because I set the regular state to `opacity .2`, which will cascade down unless you change it in the hover state. I didn't know of any other way to set opacity on a token.

andresgalante: @mrewers I wouldn't make it black, just keep the original color and change the opacity.

```css
&:hover { opacity: .5; }
```

@francocorreasosa, not related to this PR, but for later refactoring I think opacities values should be tokens.

mrewers: @andresgalante Will do. Thanks.

francocorreasosa:
Hey! Can you rebase your PR from our current `master`? So we can get it run from the CI and merge it. Thanks

francocorreasosa:
Merging this into an internal branch so we can review with a staging URL :) Thanks!

Issue #883, opened 2018-09-03
Docs: Navigation isn't responsive


siddharthkp:
<img width="788" alt="screen shot 2018-09-03 at 8 30 41 pm" src="https://user-images.githubusercontent.com/1863771/44993459-4a49e480-afb8-11e8-8f37-5c98cb56460f.png">

adamjang:
I'll take a look at this.

lopno:
I made a PR to fix this.
I was not sure if @adamjang was still working on it.

adamjang:
I have a solution which uses a hamburger menu similar to the auth0 website, but I won't be able to make the PR until tonight.

PR #965, opened 2018-10-03
Made documentation header responsive

lopno:
resp_header

Hiding the texts and leaving just the icons as the width gets smaller.
Set a min-width of 400px, since there is no way to fit everything without making a completely new design.

Fixes https://github.com/auth0/cosmos/issues/883

### DIFF
--- a/internal/docs/docs-components/navigation.js
+++ b/internal/docs/docs-components/navigation.js
@@ -17,6 +17,7 @@ const Navigation = styled.nav`
   background-color: #222;
   height: 80px;
   padding: 0 ${spacing.medium};
+  min-width: 400px;
 
   display: flex;
   flex-wrap: nowrap;
@@ -47,6 +48,9 @@ const Navigation = styled.nav`
   ul > li {
     display: inline-block;
     margin-left: ${spacing.large};
+    @media (max-width: 880px) {
+      margin-left: ${spacing.xsmall};
+    }
   }
 `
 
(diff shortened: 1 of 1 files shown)


internal/docs/docs-components/navigation.js
```diff
@@ -59,6 +63,9 @@ const LogoName = styled.h1`
   margin-left: 16px;
   margin-right: 0.75em;
   text-transform: uppercase;
+  @media (max-width: 480px) {
+    font-size: 0;
```
francocorreasosa: Could we use a `display: none;` here instead?

internal/docs/docs-components/navigation.js
```diff
@@ -67,6 +74,12 @@ const LogoContainer = styled.div`
   align-items: center;
 `
 
+const ResponsiveSpan = styled.span`
+  @media (max-width: 880px) {
+    font-size: 0;
```
francocorreasosa: Same question about the`display: none;`

francocorreasosa:
Thanks for your contrinution! Left a comment about the text hiding on small viewports.

lopno:
That was my original idea, but it seems like it is being overwritten.
image

siddharthkp:
@francocorreasosa Do you think we should implement both this and https://github.com/auth0/cosmos/pull/1013?

francocorreasosa:
@siddharthkp I think so, let's see what @landitus think about this.

lopno:
@landitus No worries! I saw the other solution, and it looked great :)

Issue #972, opened 2018-10-05
[API Consistency] Component-as-props API consistency


francocorreasosa:
# Goal 
The goal here is to have consistency in the way components are passed as props to other components, also to improve maintainability Cosmos.

For example: 

## Icon
Icon prop in Button:
```js
<Button icon={<Icon name="icon-name" />} ... />
```
Icon prop in Alert:
```js
<Alert icon={<Icon name="icon-name" />} ... />
```
or
```js
<Alert icon appearance="warning" ... />
```
for the alert to infer what icon to use.

## Avatar
We may want to keep this one as-is, since there are no other use cases for tha avatar to be passed in as a prop to another component.
```js
<AvatarBlock avatar={<Avatar image="..." type="..." ... />} ... />
```

## Actions
Empty State Action:
```js
<EmptyState action={<Button ... />} ... />
```

Form Actions:
```js
<FormActions 
  primaryAction={<Button ... />} 
  secondaryAction={<Button ... />}
  destructiveAction={<Button ... />} 
  ... 
/>
```

Page Header:
```js
<PageHeader
  primaryAction={<Button ... />} 
  secondaryAction={<Button ... />} 
  ... 
/>
```
# Checklist

- [ ] Alert (icon) #1071 
- [ ] Button (icon)
- [ ] AvatarBlock (type, image from Avatar)
- [x] Dialog (Buttons)
- [ ] Empty State (action)
- [ ] Form Actions (Buttons)
- [ ] Page Header (actions)
- [ ] Resource List (actions)

⚠️ **Important:** We need **not to break backwards compatibility** with current approaches. Current API should be deprecated until 1.0 instead.

For reference: https://twitter.com/jamiebuilds/status/1047968551758323712

sanketsingh24:
Hey! I would like to work on this issue, but would need some more info. Like,for example, We need to migrate "label"  in "Button" component to props? and so on for other components?

francocorreasosa:
@sanketsingh24 The `label` prop in button refers to the Tooltip that is shown when the button is icon only. In this particular case the `<Tooltip />` component should be used I guess. But this particular use case is not confirmed so don't worry :) Thanks anyway!

sanketsingh24:
Thanks for the help, I will look more into this and submit a PR tomorrow :D

siddharthkp:
Hi! I'm not sure if all components would benefit from this

Button is a great example because we would like to control the position/color/etc based on the `appearance` of the button 

[...]

PR #974, opened 2018-10-06
Allow icon component instance as prop in Alert

cangoektas:
Hey 👋 

I saw #972 and hope this goes into the right direction. This PR updates the `Alert` component to also accept a React element as an `icon` prop. This is part of the migration to support component instances as props.

### Questions:
- ~Should there be a validation to only accept instances of `Icon`?~
- Should usage of the string `icon` prop be deprecated?

### Todo's
- [ ] Update `Alert` docs

### DIFF
--- a/core/components/atoms/alert/alert.js
+++ b/core/components/atoms/alert/alert.js
@@ -148,8 +148,8 @@ Alert.propTypes = {
   /** Style of alert to show */
   type: PropTypes.oneOf(['default', 'information', 'success', 'warning', 'danger']).isRequired,
 
-  /** Name of icon */
-  icon: PropTypes.oneOfType([PropTypes.oneOf(__ICONNAMES__), PropTypes.element]),
+  /** Icon element or name of Icon*/
+  icon: PropTypes.oneOfType([PropTypes.element, PropTypes.oneOf(__ICONNAMES__)]),
 
   /** Title text (in bold) */
   title: PropTypes.string,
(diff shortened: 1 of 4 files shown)


francocorreasosa:
Hey! Thank you for contributing to Cosmos! 

> Should there be a validation to only accept instances of Icon?

Sounds good! How would you implement it?

> Should usage of the string icon prop be deprecated?

Yes. As you said, this needs to be a deprecation and not a breaking change. We will take care of removing deprecations once we reach to a stable version.

>  - [ ] Update Alert docs

🙏 👌 🙌

francocorreasosa:
We're using `react-docgen` for that :) I will add the `don't merge` label since it seems you're still working on this, so we don't want to merge it while in progress. I can remove it when you're ready.

core/components/atoms/alert/alert.js
```diff
@@ -10,6 +10,15 @@ import { deprecate } from '../../_helpers/custom-validations'
 import Automation from '../../_helpers/automation-attribute'
 import Icon, { __ICONNAMES__ } from '../icon'
 
+const createIconForAlert = (icon, color) => {
```
sanketsingh24: Why not do this in the "icon" component itself? It might break some components but will make the "icon" component universal and we will not have to do this in future when adding more components?

francocorreasosa: I think this could be a helper that components requiring icons as a prop can consume.

sanketsingh24: Yeah, I think that's better. Won't break anything, and won't be required to do again.

cangoektas: @francocorreasosa just opened #979

francocorreasosa:
Thank you all! We're discussing internally some things we (as the project) are missing in order to merge all your PRs as soon as possible.Thank you again for contributing!

francocorreasosa:
Closing in favor of #1071

PR #982, opened 2018-10-09
Add example for actions in table

### LINKED ISSUE #980: Add example of Table with actions in documentation
It's not clear how a table can handle actions, and given we use the Button group to align buttons, it's worth adding this example in the docs.

### Inline row actions
Inline actions allow the user to quick launch actions to the row item. This is useful to trigger editing or destructive actions from the table. A maximum of 2 actions are allowed per row, but when more actions are needed you should use the More actions button with a dropdown. 

We could add an example like this, with better data.

```
<Table
      items={[
        {
          name: '1'
        },
        {
          name: '2'
        },
        {
          name: '3'
        },
        {
          name: '4'
        }
      ]}
    >
      <Table.Column field="name" width="40%" title="Name" />
      <Table.Column field="name" width="40%" title="Description" />
      <Table.Column>
        {item => (
          <ButtonGroup align="right">
            <Button icon="pencil" />
            <Button icon="delete" />
          </ButtonGroup>
        )}
      </Table.Column>
    </Table> 
```

francocorreasosa:
Fixes #980 

image
image

### DIFF
--- a/core/components/molecules/table/table.md
+++ b/core/components/molecules/table/table.md
@@ -282,26 +282,20 @@ class TableWithActions extends React.Component {
         items={[
           {
             name: 'Management API',
-            base_url: 'https://management.example.com/api/v2/',
-            live: true
+            base_url: 'https://management.example.com/api/v2/'
           },
           {
             name: 'Authentication API',
-            base_url: 'https://id.example.com/',
-            live: true
+            base_url: 'https://id.example.com/'
           },
           {
             name: 'Feature flags API',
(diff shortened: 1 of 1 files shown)


landitus:
great ! 

I'm not sure about the 'live' column. Certainly feels like something we should think about more in design. Like an icon badge or something. I have the feeling we have something in all Auth0 materials, like a green circle with the check inside, for example. But, let's remove that column for the example, for now until we figure it out :)

francocorreasosa:
No problem, done!

PR #985, opened 2018-10-10
Avatar Fixes

francocorreasosa:
### Avatar
- [x] Fix min width to prevent breaking on small containers
- [x] Add an empty state when no image is set. There is a design of the person icon in gray for this. 
- [x] Docs: Image used in the example should be squared to be allowed to be used in both types: resource/user
- [x] Docs: Missing examples when we create a default user avatar (using initials)
- [x] Docs: For type User, using random people headshots as examples makes it more evident
- [x] Tests: Missing tests and examples when type user and no image is defined ([already done](https://auth0-cosmos.now.sh/sandbox/?selectedKind=AvatarBlock&selectedStory=defaults&full=0&addons=1&stories=1&panelRight=0))
- [x] Tests: Missing tests and examples when type resource and no image is defined ([already done](https://auth0-cosmos.now.sh/sandbox/?selectedKind=AvatarBlock&selectedStory=defaults&full=0&addons=1&stories=1&panelRight=0))

### Avatar Block
- [x] Reduce avatar block text left spacing
- [x] Type prop should have a default value of user
- [x] The image should be a link when the link prop is active
image prop should not be required as we can have default/empty states for the image (inherited by Avatar)
- [x] Docs: Documentation is lacking examples of permutations (mostly for type Resource)
- [x] Docs: For type User, using random people headshots as examples makes it more evident. Including the main example in the Props Table.


landitus:
Let's replace the Cosmos avatar with a square Cosmos avatar just to make sure it works on any Avatar type (square and rounded)

PR #991, opened 2018-10-16
Move fonts to a different package

siddharthkp:
- [x] Step 1: Move to a different package inside cosmos repo
- [x] Step 2: Move font package to a different repo [auth0/cosmos-fonts](https://github.com/auth0/cosmos-fonts)
- [x] Step 3: Use new package instead of local package
- [x] Step 4: Add link to fonts repo in usage


siddharthkp:
Have to wait until a jenkins pipeline is setup for https://github.com/auth0/cosmos-fonts

core/cosmos-fonts/index.js
```diff
@@ -0,0 +1,61 @@
+import { fonts, misc } from '@auth0/cosmos-tokens'
+
+const insertAtTheStart = styles => {
+  let tag = document.getElementById('cosmos-fonts')
```
francocorreasosa: In the server side we'd get an `undefined has not property 'getElementById'` kind error because `document` is not a thing there.

siddharthkp: should we just add a typeof undefined check?

francocorreasosa: that would do it, yeah. we would still need to find a way to apply global styles when cosmos is rendered in the server (maybe with `injectGlobal`?)

siddharthkp:
Copied the bug over to https://github.com/auth0/cosmos-fonts/issues/1

PR #1026, opened 2018-10-25
Reset line height in story wrapper

siddharthkp:
`[class^="sc-"] { line-height: 1.6; }` also applies to the story wrapper, which means any components inheriting line-height will take the wrapper's line-height. 

This makes our tests less confident as the components might not behave the same way in a different environment

Overriding this to `inherit`


siddharthkp:
Okay, bad time to do this.

Should tackle after https://github.com/auth0/cosmos/pull/955

PR #1019, opened 2018-10-24
Sebastian/button v2

jvelezpo:
Now a Button with only an icon wont change its color to a blue one

fixes: #841 

I messed up with the rebase on https://github.com/auth0/cosmos/pull/967 so i am opening this new PR and closing the old one, I also have other 2 PR open, can someone help me so i dont have to do the same for those others PR, thanks.

PR #1033, opened 2018-10-25
Button with icon only color fix via @jvelezpo

### LINKED ISSUE #841: Button with icon only: Icon will be blue, not button's text color
The button component forces a special size and color when the content is an icon:

/core/components/atoms/button/button.js@master#L158-L163

In cases like the close btn on a dialog and others, this is not desired. I suggest removing this declaration, and add a new appearance for "action" btns that looks like regular text.

Therefore we can create things like kebab menus or close btns.

Before resolving this issue please make sure the overwrite on the dialog component for the close btn is removed.

francocorreasosa:
Continues #1019

Fixes #841

### DIFF
--- a/core/components/atoms/button/button.js
+++ b/core/components/atoms/button/button.js
@@ -159,7 +159,6 @@ const getAttributes = props => {
   if (props.icon && !props.text) {
     styles.padding = 0
     styles.minWidth = '36px'
-    styles.icon = colors.button.link.icon
   }
 
   return styles


siddharthkp:
@landitus Blocked on your review

landitus:
This PR is related to https://github.com/auth0/cosmos/pull/1251 and we'll resume work on this once we design what option is best from here: https://github.com/auth0/cosmos/pull/1251#issuecomment-447919781

Sorry @jvelezpo 	that we can't move forward with this one :)

Issue #1004, opened 2018-10-22
Add a subtitle to the Page Header component


landitus:
Add a subtitle to the Page Header component

MiniMarvin:
I think I may work on it

PR #1038, opened 2018-10-27
feat: Added subtitle to page-header component

MiniMarvin:
Now the component from page header has an optional subtitle option which
creates a gray text next to the component name. As expected in the
contribution guideline the story to this new feature is now on.

Resolves: #1004

### DIFF
--- a/core/components/molecules/page-header/page-header.js
+++ b/core/components/molecules/page-header/page-header.js
@@ -26,6 +26,13 @@ const StyledPageHeader = styled.div`
   }
 `
 
+const StyledSubtitle = styled.div`
+  color: gray;
+  margin-top: 0;
+  margin-bottom: -12px;
+  padding-left: ${spacing.xxsmall};
+`
+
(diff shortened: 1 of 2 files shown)


core/components/molecules/page-header/page-header.js
```diff
@@ -26,6 +26,13 @@ const StyledPageHeader = styled.div`
   }
 `
 
+const StyledSubtitle = styled.div`
+  color: gray;
```
francocorreasosa: We can use tokens here

core/components/molecules/page-header/page-header.js
```diff
@@ -26,6 +26,13 @@ const StyledPageHeader = styled.div`
...
+  margin-top: 0;
+  margin-bottom: -12px;
```
francocorreasosa: 🤔 We do we need a negative margin here?

francocorreasosa:
Merging this into `1038-added-subtitle-to-page-header` so we can continue there.

PR #1045, opened 2018-10-30
WIP: Replace sc hack with container styles Part 2

siddharthkp:
Carrying #955 forward

### DIFF
--- a/core/components/atoms/breadcrumb/breadcrumb.js
+++ b/core/components/atoms/breadcrumb/breadcrumb.js
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types'
 import Automation from '../../_helpers/automation-attribute'
 
 import { fonts, spacing, colors } from '@auth0/cosmos-tokens'
+import containerStyles from '../../_helpers/container-styles'
 
 import Icon, { StyledIcon } from '../icon'
 import Link, { StyledLink } from '../link'
@@ -17,8 +18,9 @@ const LinkIcon = styled(Icon)`
 `
 
 const Wrapper = styled.div`
-  ${/* overrides for link */ ''};
+  ${containerStyles};
 
+  ${/* overrides for link */ ''};
   ${StyledLink} {
     color: ${colors.base.grayDarkest};
     font-size: 13px;
--- a/core/components/atoms/button/button.js
+++ b/core/components/atoms/button/button.js
@@ -107,21 +107,25 @@ const states = {
 
 const sizes = {
   large: {
+    height: misc.button.large.height,
     lineHeight: subtract(misc.button.large.height, '2px'),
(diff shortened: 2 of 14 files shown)


siddharthkp:
@landitus Can you check the Dialog snapshot. I've removed the 2px adjustment from master, can't say if the dialog looks better or worse: https://www.chromaticqa.com/snapshot?appId=5aba7f6a07f95800202658ff&id=5bdbd8fd0808790024e04bfe

andresgalante:
@siddharthkp as far as I can tell, this PR create a variable called `containerStyles` that onnly defines a linne height of 1.6, and it applyes it to each component that needs a 1.6 line height. If that's the case, then:

1- Yay! I am learnig react, and
2- It's the way to go. On one hand, it's less performant but on the other, it shields the components. It's hackier than the prefix solution IMO, but it works and it's temporary.
(If I am completely missing the point, please let's have a talk, it'll save a lot of time.)

Using a span inside the alert seems odd to me, it's not semantic. But I can see how the paragraph having margin-top by default is breaking the alert. When we deal with margins lets fix this.

Nothing should ever have a line-height of 1, this is a red flag. We should clean it across the system. Probably a task to do when we create a token system.

When we create new components let's make sure to remind ourselves to use this variable.

One idea: for users that are not using cosmos with the styled guide side by side, can we provide a way for them to build cosmos without this option?

core/components/atoms/button/button.js
```diff
@@ -268,10 +268,10 @@ Button.Element = styled.button`
 Button.Text = styled.span`
   display: inline-block;
   vertical-align: middle;
-
+  line-height: normal;
   /* Sub-pixel position adjustment */
-  /* See: https://github.com/auth0/cosmos/pull/947 */
-  margin-top: 2px;
+  /* See: pulls 947 and 1045 */
```
francocorreasosa: Can we include full URLs to the PRs? So in most editors you can `Cmd + Click` it to enter the link.

francocorreasosa:
```suggestion
  /* See: https://github.com/auth0/cosmos/pull/947 */
  /* See: https://github.com/auth0/cosmos/pull/1045 */
```

core/components/atoms/breadcrumb/breadcrumb.js
```diff
@@ -17,8 +18,9 @@ const LinkIcon = styled(Icon)`
 `
 
 const Wrapper = styled.div`
-  ${/* overrides for link */ ''};
+  ${containerStyles};
 
+  ${/* overrides for link */ ''};
```
francocorreasosa: Why do we need this empty string to be inserted here?
```suggestion
  /* overrides for link */
```

PR #1050, opened 2018-10-30
Hide pagination toolbar if 1 or zero pages (or zero items)

### LINKED ISSUE #1049: PaginationToolbar - Hide if total === 0
PaginationToolbar will hide itself if  `items` is <= the `perPage` attribute. 

This should also happen if items == 0

image

astanciu:
Fixes #1049


siddharthkp:
Good catch!

I'm merging this to a local branch, I'll add some tests and push in the next release 🙌

PR #961, opened 2018-10-02
Add accesibility to Tabs

jvelezpo:
The Tabs can now be navigated using the keyboard by hitting tab

Fixes: https://github.com/auth0/cosmos/issues/923

[x] tests passes

PR #1052, opened 2018-11-01
Make tabs accessible 

siddharthkp:
Copied changes from #961


francocorreasosa:
I'll update the snapshots then we can merge

PR #1037, opened 2018-10-27
[Table] Add accessibility to table header and table row.

rfbotto:
Closes https://github.com/auth0/cosmos/issues/872

PR #1057, opened 2018-11-02
[Table] Add accessibility to table header and table row. (#1037)

### LINKED ISSUE #872: Table component is not keyboard accessible
Sortable table header cells and clickable table rows cannot be interacted with via a keyboard, as they do not have the necessary `tabindex` attribute.

siddharthkp:
#1037 by @rfbotto 

Fixes #872

### DIFF
--- a/core/components/molecules/table/table-header.js
+++ b/core/components/molecules/table/table-header.js
@@ -18,7 +18,7 @@ const TableHeader = props => {
       )
     }
 
-    const onClick = column => {
+    const sortColumn = column => {
       if (!column.sortable) return
 
       let sortDirection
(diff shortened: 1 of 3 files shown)


siddharthkp:
> is it worth just using a third party library?

@andresgalante I'd really like that, something lightweight that we can skin

siddharthkp:
@andresgalante Does this PR makes things worse to manage?

If not, we can merge this and replace the whole thing with a library when we need to add features to table

siddharthkp:
Not sure 🤔 

This PR is super old, copied changes to new https://github.com/auth0/cosmos/pull/1161

PR #1059, opened 2018-11-02
[RFC - Accepted] margin prop

siddharthkp:
### Problem statement

Cosmos takes care of the _design_ of a component, ideally an application using cosmos should not modify the component's styles

There is one big exception, positioning. The application has to decide how to position it relative to other components.

Some of these concerns can (should?) be solved in cosmos with *layout components*. Example: The spacing between 2 consecutive paragraphs should not be the developer's concern

But, there are other cases when the application needs to position the component through margins.

Here's an example:

```jsx
const Page = props => (
  <ApplicationHeader >
    <Heading size={1}>{props.title}</Heading>
	<Heading size={3}>{props.description}</Heading>
  </ApplicationHeader>
)

const ApplicationHeader = styled.div`
  /* application code */
  margin-bottom: ${spacing.large};
  
   /* override cosmos component by tagname */
   h1 {
     margin-bottom: 8px;
   }

   /* or by reference */
   ${StyledHeading[3]} {
     margin-bottom: 16px;
   }
`
```

Overriding components like this creates a few long term problems

1. It's an open API that promotes bad practice.

   _New team member: "Oh cool, this is how I can override styles. I'll just add a color property here"_

2. It makes the application fragile to updates

   One of the goals of cosmos is to improve the design of applications without requiring giant redesign efforts. Because this approach depends on either the rendered output or the internal implementation details, it's fragile and hard to track + test

&nbsp;

_Note: This could be handled in the application without modifying the component by creating a wrapper around it. But it doesn't really solve problem 2 from above and introduces unnecessary wrapper elements in the output (bad for responsiveness and accessibility)_

Example:
```js
const Page = props => (
  <ApplicationHeader >
    <div style={{ marginBottom: 8 }}>
      <Heading size={1}>{props.title}</Heading>
    </div>
  </ApplicationHeader>
)
```

&nbsp;

# Suggested solution

Creating an explicit prop API that makes it easy to 

1. Predictable
2. Easier track + test

&nbsp;

```diff
const Page = props => (
  <ApplicationHeader >
-   <Heading size={1}>{props.title}</Heading>
+   <Heading size={1} margin={{ bottom: 8 }}>{props.title}</Heading>

-   <Heading size={3}>{props.description}</Heading>
+	<Heading size={3} margin={{ bottom: 'small' }}>{props.description}</Heading>
  </ApplicationHeader>
)

const ApplicationHeader = styled.div`
  /* application code */
  margin-bottom: ${spacing.large};
  
-   /* override cosmos component */
-   h1 {
-     margin-bottom: 8px;
-   }

-   ${StyledHeading[3]} {
-     margin-bottom: ${spacing.small}
-   }
`
```

The `margin` prop can be passed to any cosmos component. The values can be any of the spacing tokens or an exact value like `1em` and `8` (assumes px by default)

&nbsp;

This solution is inspired from [styled-system](https://github.com/jxnblk/styled-system) and suggestions from other [design system developers](https://twitter.com/siddharthkp/status/1042780297744539649)

_Note: There are cases where we allow overriding styles other than margin, they suffer from the same problems mentioned above. This proposal does not address those, but we will come back to them in the coming weeks_

&nbsp;

### Implementation detail (for cosmos contributors)

A helper that can be imported and used in any component component

```jsx
import { colors, fonts } from '@auth0/cosmos-tokens'
import margin from '../../_helpers/margin'

const BaseHeading = styled.h1`
  margin: 1em 0;
  ${margin}; /* overrides */
  
  color: ${colors.text.default};
  font-weight: ${fonts.weight.normal};
`
```

The order is important here. `${margin}` should always come after the margin defined in the component

Look at the `Files changed` of this PR to see it in action

#### Room for improvement

I'm not a big fan of adding boilerplate to the component because we will also have to enforce it as a rule and some one (me) will always forget to add it.

We can add it to the component/code review checklist or add a linter for this, but a better solution would be to abstract it behind a component creation API

```jsx
import component from '../../_helpers/component'
import { colors, fonts } from '@auth0/cosmos-tokens'


const BaseHeading = component.h1`
  margin: 1em 0;
  color: ${colors.text.default};
  font-weight: ${fonts.weight.normal};
`
```

which is equivalent to:

```jsx
import { colors, fonts } from '@auth0/cosmos-tokens'
import margin from '../../_helpers/margin'

const BaseHeading = styled.h1`
  /* component styles */
  margin: 1em 0;  
  color: ${colors.text.default};
  font-weight: ${fonts.weight.normal};

  /* override utilities */
  ${margin}
`
```

&nbsp;

#### Overriding styles for these cases needs some more thinking

1. Compound components

    Example from manage:

    ```jsx
    FUDPlatform.Title = styled(StyledHeading[4])`
      margin: 10px 0;
    `
    ```

    can be refactoring to the following without losing any details (I think)

    ```jsx
    FUDPlatform.Title = (props) => (
      <Heading size={4} margin={{ top: 10, bottom: 10 }} {...props}/>
    )
    ```

2. Layout components (inside cosmos)

    I'd love to follow this approach inside cosmos as well, but we don't always know the  children/DOM before hand

    ```jsx
    const ButtonGroup = styled.div`
      display: flex;
      justify-content: center;

      /* All immediate children (buttons) should have a tiny margin between them */
      & > * {
        margin: ${spacing.xsmall};
      }
    `
    ```

    Not sure what's the best way to keep this clean

    The `Stack` component uses a slightly annoying method of wrapping each child in a `StackedItem` component to handle margin (and width)

    ```jsx
    render(
      <Stack>
        <TextInput/>
        <Button>Save</Button>
      <Stack>
    )
    ```

    results in the following not pretty DOM output:

    ```html
    <div data-cosmos-key="stack">
      <div data-cosmos-key="stack.item">
        <input data-cosmos-key="text-input" type="text" value="">
      </div>
      <div data-cosmos-key="stack.item">
        <button data-cosmos-key="button">
          <span>Save</span>
        </button>
      </div>
    </div>
    ```


andresgalante:
Hi @siddharthkp, Thanks a lot for starting this discussion here in the open, it's super to define this sooner than later. I'd like to understand this better. If it's ok with you let's talk next week. Here is my gut reaction:

I can absolutely see the need for some components having the option to opt-in a margin prop, for example on headings. In those cases, I'd be extremely careful to only open a subset of spacer coming from our global variable. Allowing the users to introduce magic numbers is a disaster to consistency.

Keep in mind that nothing should have a top margin.

On the other hand, I'd avoid doing it for every component. Let's be really careful to what we allow since the moment an author adds a margin, that component is no longer modular, creating a maintainability issue down the road.

It's always easier to see than to tell, an idea I have is to quickly mock an POC of what the output would be of different layouts and components working together. There we can spot any problems in the no margin approach. What do you think?

siddharthkp:
I think you meant to tag @landitus :)

core/components/_helpers/margin.js
```diff
@@ -0,0 +1,33 @@
+import { spacing } from '@auth0/cosmos-tokens'
+
+const directions = ['top', 'bottom', 'left', 'right']
```
francocorreasosa: I agree with @andresgalante, we shouldn't allow `top` margin.

francocorreasosa: I agree with @andresgalante, we shouldn't allow top margin.

andresgalante:
Here is a thought I have: even though things should  not a margins top, by allowing it as a helper it allows us to remove all margins and offer them as optin.

Therefore we will break user that want to migrate to the next version but we offer an out of the box solutions to mimic the result they had. Does it make sense? What do you think?

core/components/atoms/heading/heading.js
```diff
@@ -2,9 +2,11 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import { colors, fonts } from '@auth0/cosmos-tokens'
+import margin from '../../_helpers/margin'
 
 const BaseHeading = styled.h1`
   margin: 1em 0;
```
andresgalante: Since we are offering it as an opt-in, can we remove this default margin?

siddharthkp:
I hadn't talked about one aspect of this solution

Adoption/Migration:

Right now cosmos components have top margin as well, this isn't what we want long term though. We would rather create some Layout components that introduce default spacing (bottom) between components.

The margin API suggested in this proposal solves the use case that we have today - which means we need to let our users override top margins as well. We can keep this closed - you can only set it to 0

We can deprecate this redundant value when it's not needed anymore.

Alternatively, we can also park this solution until we create Layout components and remove top-margins from all components

Either path leads to the same result.

core/components/_helpers/margin.js
siddharthkp: I'm not sure if we can do that right now. See thread: https://github.com/auth0/cosmos/pull/1059#issuecomment-435839056

core/components/atoms/heading/heading.js
siddharthkp: We can remove top margin as long as we give them an easy to keep their visual styles

Either with this margin API or by creating smart layout components which bake in the defaults (cc @landitus)

siddharthkp:
Bonus: We can tokenise margin properties as a group

Example: 

```js
const PageLayoutItemMargin = [0, spacing.small, spacing.large, spacing.small]

/* can be used as styles */
const PageLayout = styled.div`
  > * {
    margin: ${...PageLayoutItemMargin}
  }
`

/* and be used as a prop */
render(
  <PageHeader margin={PageLayoutItemMargin} />
)

```

andresgalante:
👍 I park it, let's just merge this and work on layouts on the side. We need both.

As you mention not remove anything so we don't create a breaking change, add the layouts and build manage example with layouts, components, and margins when needed.

@francocorreasosa @landitus @siddharthkp On v1 we remove the default margins on components to follow the system's rules and have them only as opt-in, sounds like a plan?

siddharthkp:
Sounds like a plan. Closing this PR - decisions made ✨ 

The proposal does not be merged, we will bring this back after we have layouts 👌

core/components/atoms/heading/heading.js
```diff
@@ -2,9 +2,11 @@ import React from 'react'
...
+  ${margin}; /* overrides */
```
siddharthkp: This is where the magic happens

Saifadin:
I like this approach and it gives a lot of declarative power to adding margins. 

For us it might help us correct our button styles if we combine it with `defaultProps` to migrate slowly and fix the stylings.

On the other hand this does not pose a solution for media-query based margins, for example smaller margins on mobile then on desktop. One way is of course using something like `react-media` or a `useMedia` hook, when React Hooks comes out. But that would add a lot of overhead and move styling into the JS part, which might not be desirable.

What other solutions can you imagine?

Saifadin:
So a few take-aways:

Yesterday I implemented this in our internal design-system and I find it to be really good. For now I only implemented it with our `Button` component and I found it to be practical. There was only one use-case where I needed a `@media` query and I did not find it problematic to to something like this:
```jsx
export const FooterButton = styled(Button)`
  @media (min-width: 600px) {
    margin-bottom: ${ spacing * 2 };
  }
`;

<FooterButton margin={{ bottom: 1 }}>...</FooterButton>
```
Because we are only using multiples of 8px in our system the numbers passed to `margin` are multiplications, would love to hear your opinion on that 🤔

Also I think it is a valid usecase for things like `ButtonGroup` or `Layout` to have overrides through their css. I never found myself wanting to override them. After all, that is part of their functionality in my opinion.
```jsx
export const ButtonGroup = styled.div`
  > * {
    margin-right: ${spacing * 2};
    &:last-child {
      margin-right: 0;
    }
  }
`;
```

PR #1061, opened 2018-11-02
Restore pagination button selected state

francocorreasosa:
image


landitus:
should we trigger another build @francocorreasosa ?

francocorreasosa:
Actually we have to rebase from master for the new structure to come here. Will do after finishing Select fixes.

PR #1064, opened 2018-11-04
Disabled option in select component

rakhi2104:
### What's working (Resolves #1062 )
- Added the feature to identify the option if it is readOnly.
### What's not working
- Examples of `select` component fail to work in **readOnly** mode in Stories (reference [here](https://github.com/auth0/cosmos/issues/1062#issuecomment-435575655))

PR #1067, opened 2018-11-05
Select: Add disabled + disabled options + groups

### LINKED ISSUE #1062: Add disabled option to Select
We are missing a way to disabled options inside the Select component. This is a native feature to HTML.

francocorreasosa:
Extends #1064 
Resolves #1062

### DIFF
--- a/core/components/atoms/select/select.js
+++ b/core/components/atoms/select/select.js
@@ -6,19 +6,15 @@ import { css } from 'styled-components'
 import { misc } from '@auth0/cosmos-tokens'
 import { StyledInput } from '../_styled-input'
 
-const disabledOpacity = 0.5
+const selectOpacity = {
+  default: 1,
+  disabled: 0.6
+}
 const PLACEHOLDER_VALUE = '__select_placeholder'
-
-const disabledState = props =>
-  props.disabled
-    ? css`
-        opacity: ${disabledOpacity};
-      `
-    : null
 
 const StyledSelect = StyledInput.withComponent('select').extend`
   height: ${misc.input.default.height};
-  ${disabledState}
+  opacity: ${props => (props.disabled ? selectOpacity.disabled : selectOpacity.default)};
 `
 
 const Select = ({ options, ...props }) => {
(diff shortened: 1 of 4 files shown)


core/components/atoms/select/select.js
```diff
@@ -20,14 +22,14 @@ const Select = ({ options, ...props }) => {
...
       {/* First option will be selected if there is no value passed as a prop */}
-      <option disabled hidden selected={!props.value} value="" {...Automation('select.option')}>
+      <StyledOption disabled hidden selected={!props.value} value="" {...Automation('select.option')}>
         {props.placeholder}
-      </option>
+      </StyledOption>
 
       {options.map((option, index) => (
-        <option key={index} value={option.value} {...Automation('select.option')}>
+        <StyledOption key={index} value={option.value} readOnly={option.readOnly} disabled={option.readOnly} {...Automation('select.option')}>
```
siddharthkp: Err why?

rakhi2104: the readOnly styling can be implemented this way. Tried css selectors, but didn't work.

francocorreasosa: We can just use the `disabled` attribute of the `<select />`

rakhi2104: [here](https://github.com/auth0/cosmos/issues/1062#issuecomment-435567902) I discussed the hardly noticeable difference of the disabled select option

siddharthkp: We need to use the disabled attribute either way for accessibility, I think we can add our custom styles to make it look more disabled 🙂

francocorreasosa: Fixed it :)

francocorreasosa:
@landitus can you take a look whenever you have time?

landitus:
I'm attaching the states for reference

screen shot 2018-11-06 at 10 53 19

francocorreasosa:
Fixed, and added reference to the W3C in the docs :)

landitus:
Let's make the read-only opacity 1, so it's just a native read-only select

landitus:
Sorry about the confusion @francocorreasosa, as Selects don't have readonly, only disabled. 

Here's the updated design. The disabled opacity is 0.5.

francocorreasosa:
No problem @landitus. Just updated the code to reflect that, thanks.

PR #1068, opened 2018-11-06
Add viewports for new chromatic feature

siddharthkp:
Ref: http://docs.chromaticqa.com/viewports


siddharthkp:
storybook 4 uses babel 7 while cosmos-scripts uses babel 6, will start there

PR #1069, opened 2018-11-06
Internal: Upgrade to babel 7 and storybook 4.0.4

siddharthkp:


siddharthkp:
Ugh, babel-jest and babel 7 are not nicely compatible though. Solutions in the issues recommend using alpha and bridge versions, I'm going to park this until I understand how they work

PR #1071, opened 2018-11-07
[API Consistency] Alert: Icon component as prop

francocorreasosa:
Deprecates:
```js
<Alert icon="warning" ... />
```

in favor of:
```js
<Alert icon={<Icon name="warning" />} ... />
```

### DIFF
--- a/core/components/_helpers/icon-prop.js
+++ b/core/components/_helpers/icon-prop.js
@@ -1,0 +1,12 @@
+import React from 'react'
+import Icon from '../atoms/icon'
+
+export default prop => {
+  if (typeof prop === 'string') {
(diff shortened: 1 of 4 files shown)


siddharthkp:
Same comments as https://github.com/auth0/cosmos/pull/1074#issuecomment-436923399

landitus:
I think it would make sense to add some Icon defaults according to the appearance. This would fit most of the use cases and still, the user would be able to override it. So, in our case it would be the following:

- default: notes
- information: megaphone
- success: check-circle
- warning: warning
- danger: danger

screen shot 2018-11-08 at 20 09 22

-

francocorreasosa:
Closing this till we have well defined how we want to do it.

PR #1074, opened 2018-11-07
[API Consistency] Button: Icon component as prop

francocorreasosa:
**⚠️ Do not merge before #1071.**

Deprecates:
```js
<Button icon="warning" ... />
```

in favor of:
```js
<Button icon={<Icon name="warning" />} ... />
```

### DIFF
--- a/core/components/_helpers/icon-prop.js
+++ b/core/components/_helpers/icon-prop.js
@@ -1,0 +1,12 @@
+import React from 'react'
+import Icon from '../atoms/icon'
+
+export default prop => {
+  if (typeof prop === 'string') {
+    console.warn('Passing icons as string has been deprecated and will be removed in Cosmos 1.0.')
+    console.log(`Please replace 'icon="${prop}"' with 'icon={<Icon name="${prop}"/>}'`)
+    return <Icon name={prop} />
+  }
+
+  return prop
+}
--- a/core/components/_helpers/icon-prop.js
+++ b/core/components/_helpers/icon-prop.js
@@ -1,11 +1,19 @@
 import React from 'react'
 import Icon from '../atoms/icon'
 
-export default prop => {
+const applyAdditionalProps = (element, props) => {
+  const propsArePresent = props && Object.keys(props).length > 0
+
+  return propsArePresent ? React.cloneElement(element, props) : element
+}
+
+export default (prop, additionalProps = {}) => {
   if (typeof prop === 'string') {
     console.warn('Passing icons as string has been deprecated and will be removed in Cosmos 1.0.')
     console.log(`Please replace 'icon="${prop}"' with 'icon={<Icon name="${prop}"/>}'`)
-    return <Icon name={prop} />
+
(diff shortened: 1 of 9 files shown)


siddharthkp:
Whoa, I have a bunch of questions here

1. Should we deprecate the icon `name` API? Icon component sounds like an opt-in for advanced functionality to me. What are some everyday use cases that can't be solved with icon name?

2. Does the icon component support size and color inside a button?

```jsx
<Button appearance="link" icon={<Icon name="dashboard" size="20" color="orange" />} />
```

2.1 Should it support all `Icon` props?
2.2 If not, is there an API design flaw here?

francocorreasosa:
No problem at all
> Should we deprecate the icon name API or support both?

I think we should support only one, supporting both would cause confusion of which one to use when.

> Does the icon component support size and color inside a button?

Those values get overriden to ensure component visual stability (see https://github.com/auth0/cosmos/pull/1074/files#diff-173e0f38f8f7a3246380686774877ad3R7)

> 2.1 Should it support all Icon props?

All except host component's overrides

> Sorry, especially for this: For rewriting API for consistency, we should probably write down similar APIs in one place before making code changes

Agree, let's do it

landitus:
Just to be clear on @siddharthkp's question, from the design perspective, we should not allow the user to set the icon color or size for buttons, as they are tied to the appearance and the system controls it (or overrides it internally.

francocorreasosa:
@landitus Exactly, we don't. See https://github.com/auth0/cosmos/pull/1074/files#diff-173e0f38f8f7a3246380686774877ad3R15 where the host's props are applied to the passed component and https://github.com/auth0/cosmos/pull/1074/files#diff-7af9b2574de0761e822dc5740731a600R173 where host's component props are being set.

siddharthkp:
Blanket statement: (Copied from [#972](https://github.com/auth0/cosmos/issues/972#issuecomment-437305318))

I don't think the goal should be to move to a component-instance-as-props API all across.

**That's a goal that starts with a solution in mind, not a problem.**

How I see it, we need to look at use-cases for each type of feature.

Example: Adding an icon should feel the same for a button and input. Adding an action should feel the same for empty state and page header.

---

> Should we deprecate the icon name API or support both?

> I think we should support only one, supporting both would cause confusion of which one to use when.

I'd recommend looking at the use cases for passing addition props to Icon

If 90% of the times, you only need name, that's the right API to keep. In cases where you need to add additional props to the Icon, you can opt-in to the more verbose syntax.

_Note: This is the opposite of what happens with actions, you always need to pass multiple props, where the verbose API would fair better_


---

> From the design perspective, we should not allow the user to set the icon color or size for buttons, as they are tied to the appearance and the system controls it (or overrides it internally)

This is where the verbose API really fails, if you want the user to pass an Icon component instead of just name, you have to respect the props they give. Chances are they really have a need to override the system. (We should fallback on system props if the user did not pass them)

francocorreasosa:
Closing this till we have well defined how we want to do it.

PR #1078, opened 2018-11-09
feat: Added subtitle to page-header component via @MiniMarvin

francocorreasosa:


siddharthkp:
Not sure if this is the right implementation, tagged @landitus on the original issue (https://github.com/auth0/cosmos/issues/1004)  After he get's back to this, we can re-open this PR

PR #1079, opened 2018-11-12
Adds system guidelines to the docs

andresgalante:
This PR defines Cosmos guidelines.

PR #1085, opened 2018-11-13
#1083 Text position change in component Switch

### LINKED ISSUE #1083: Support for text position change in Switch
The current switch has a default text position to the right of the Switch. To give more flexibility we should support moving the text to the left of the switch. This is especially useful when the switch is placed in the last column of a Table / Resource List.

screen shot 2018-11-13 at 13 18 10

Note that in the Form.Switch, the current (default) switch is maintained

.

m4tr1k:
Fixes #1083

As @landitus wanted, I developed a way to put the label on the left side if the user passes `labelLeft` as a prop, this prop is `false` by default, so the label stays on right side. One small detail, in styled-components I could not apply `${sppacing.small}` into the padding, I would like to know how to do it.

I think this is working, but I'm kind of beginner on React, but I know some things. I would like some opinions and constructive advices because I want to learn new things.

Thank you

### DIFF
--- a/core/components/atoms/switch/switch.js
+++ b/core/components/atoms/switch/switch.js
@@ -71,8 +71,8 @@ const Label = styled.label`
   letter-spacing: 1px;
   text-transform: uppercase;
   color: ${colors.text.secondary};
-  padding-left: ${props => props.labelLeft ? '0' : '16px'};
-  padding-right: ${props => props.labelLeft ? '16px' : '0'};
+  margin-left: ${props => props.labelLeft ? '0' : spacing.small};
+  margin-right: ${props => props.labelLeft ? spacing.small : '0'};
 `
 
 class Switch extends React.Component {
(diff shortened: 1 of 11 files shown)


andresgalante:
Hi @m4tr1k Thank you very much for this PR! The left label on the switch is a feature we really wanted to move forward, and having your code is fantastic.

@francocorreasosa and I have done some minor changes to reduce the code repetition on your conditional statement.

@lancelaird I labeled this PR as not merge because you when the label is on the left, the position of the switch jumps when the width of the label changes. We can't have a fixed width because we don't know how long is the string. We could potentially fix with with CSS grids (maybe) but it wont be supported by IE11.

francocorreasosa:
I think you meant to tag @landitus @andresgalante 😅

m4tr1k:
Thank you so much @andresgalante, I'm glad you liked my code 😄 Can you explain a little bit better what's causing that problem?

andresgalante:
Sure, as the size of the label changes, due to the length of the string, then it also moves the position of the switch.

The only way to solve this is by having a fixed width on the label, which is something to avoid since we don't know the size of the text the user will input there.

makes sense?

m4tr1k:
Yeah, I know what you are saying... 
One thing that can correct this issue is if `this.props.labelLeft` is true, then you move all the component to the right. With that, you can write a very long label, and the switch does not move. I will show some examples right here (I used inspect on chrome and I wrote `float: right` just to put the element on the right):
sem titulo
sem titulo1

I can't think about any other solution for that... What do you think?

andresgalante:
Although that'd be a good idea for a page, it isn't for a design system.

The position of the component is not a component concern. The component doesn't know where it is. We are building layouts to position and size components.

If you want to learn more about the separation of concerns please read this PR: https://github.com/auth0/cosmos/pull/1079

The fact that it's on the right in the case is very usecase specific and it loses modularity. Also positioning elements with float is a hack that's not been used anymore, things should be placed with flex or grid

andresgalante:
The CSS grids require some investigation. I wonder if we can potentially create 2 columns, one for the label, the other for the switch, and somehow use a relative unit to fix the width. 

But! As you mention before left align labels are probably to use this comopnent on the right. Is that's the case then there is no shift and we have no problem.

I'd wait till Monday that @landitus is around to decide. If it was for me I'd just merge this PR

m4tr1k:
Ok, if @landitus decide to put the component on the right, then I can maybe do that (I'm saying maybe because I'm on college and I don't have to much time to invest on this unfortunately 😢)

m4tr1k:
Thank you 😄

landitus:
Hey! thanks @m4tr1k :) I'm clarifying as requested: the use case for the label position option is indeed when the Switch is used flushed to the right, on a table or resource list, as seen in the image. This way, we won't have a text shift when flipping the switch. 

screen shot 2018-11-20 at 10 15 57

m4tr1k:
@landitus thanks, I will do that as soon as I come home!

m4tr1k:
Because of the `display: inline-block` now be `display: flex`, some of changes on chromatic are different, but I think that those changes don't affect the component (correct me if I'm wrong 😅 )

core/components/atoms/switch/switch.js
```diff
@@ -113,14 +124,17 @@ Switch.propTypes = {
   /** Labels to show, import for accessibility */
   accessibleLabels: PropTypes.arrayOf(PropTypes.string),
   /** Locked switch */
-  readOnly: PropTypes.bool
+  readOnly: PropTypes.bool,
+  /** Label on left side */
+  labelLeft: PropTypes.bool
```
siddharthkp: Can we change the `labelLeft` to `labelDirection: ['left', 'right']` with default as 'right'?

m4tr1k: Of course! I will try to do that now

m4tr1k:
@andresgalante Thank you for the improvements on the CSS, I think that way is more clear and I agree with those changes 😃

siddharthkp:
Merged into local branch to test on staging (https://github.com/auth0/cosmos/pull/1171)

PR #1094, opened 2018-11-19
Apply union type fix from #819

francocorreasosa:


internal/docs/spec/prop-string.js
```diff
@@ -94,12 +94,12 @@ const getPropString = propData => {
...
     */
     if (propData[name].type.name === 'union') {
-      if (propData[name].value === 'number') {
-        propString += ` ${name}={${propData[name].value}}`
-        return true
-      } else if (typeof propData[name].value === 'string' && propData[name].value.length > 0) {
+      if (typeof propData[name].value === 'string') {
         propString += ` ${name}="${propData[name].value}"`
         return true
+      } else {
```
siddharthkp: It's risky to have an else case here because we don't really know all the things that come in it.

We can handle the ones we know and leave space for the future case

(maybe you wanted `type.name === 'number'` here)

PR #1097, opened 2018-11-19
Add new logo variant

francocorreasosa:


siddharthkp:
Design looks good

I didn't understand what the API meant, are `isologo` and `isologotype` a known language?

francocorreasosa:
I still need to change those words (they come from a Spanish meaning). Maybe we can do a “full” prop? (If present then its the letters one, else is only the shield)

siddharthkp:
> they come from a Spanish meaning

Haha, that's why they felt foreign to me :D

`full` sounds reasonable

francocorreasosa:
Done ✨

PR #1098, opened 2018-11-19
Sortable Resource List

francocorreasosa:
Now, the `sortable` prop can be used to make the list sortable by dragging and dropping items. You will need to implement the `onSortEnd` method in order to reorder the items using the provided `arrayMove`

kapture 2018-12-04 at 18 22 31


francocorreasosa:
Sounds good @landitus, thank you

francocorreasosa:
Sounds like a good idea @andresgalante, let's start planning that as soon as we merge this

landitus:
@francocorreasosa 	Here's a design mockup for the Re-order feature. Let me know if there is something missing. I can walk you through it, as well.

resource list - reorder

andresgalante:
@landitus Having a tooltip to the left of the icons might be problematic. If this list is used in a place where there is no sidebar, then that tooltip would be outside the viewport.

If there any reason why the arrow icon is blue? what if we make the arrow appear only on hover?

landitus:
I agree the tooltip can be problematic. Let's put it on top.

Regarding the design of the component and the handles, I opted to leave it as is to match the Manage UI. But I have some ideas to improve it for a second iteration (yes, the handles bug me too!).

francocorreasosa:
@landitus 	I've added the `design needed` label since we need a proper design for the selected state of the RL item (for accessible sorting with the keyboard).

francocorreasosa:
@landitus using this outline design in the meantime to differentiate the selected and non-selected row items:
image

landitus:
Thanks @francocorreasosa 	! Just did a small adjustment to the focus state. You can check the updated invision prototype too.

screen shot 2018-12-10 at 15 32 22

landitus:
Just a small comment: When dragging an element over another, the tooltip is re-triggered and it creates an unpleasant effect.

I think disabling pointer events with CSS (`pointer-events: none;`) while an element is being dragged could be a good fix.

kapture 2018-12-18 at 15 13 00

francocorreasosa:
@landitus 	It should be fixed by now :)

PR #1101, opened 2018-11-20
Truncate table cell text & Add Table.Column docs

### LINKED ISSUE #857: Ability to truncate text content inside table cells to avoid multiple lines

### LINKED ISSUE #913: Add documentation for Table.Column

francocorreasosa:
Continues #916
Fixes #857
Fixes #913


core/components/molecules/table/table.js
```diff
@@ -1,12 +1,15 @@
...
-import styled from 'styled-components'
+import styled, { css } from 'styled-components'
 import { colors, spacing, misc } from '@auth0/cosmos-tokens'
 import Spinner from '../../atoms/spinner'
 import TableColumn from './table-column'
 import TableHeader from './table-header'
+
 import Automation from '../../_helpers/automation-attribute'
 
+const trucatedTableCellHeight = '45px'
```
siddharthkp: what does this do?

core/components/molecules/table/table.md
```diff
@@ -49,6 +49,17 @@
 </Table>
 ```
 
+### Table Column
```
siddharthkp: hustler :D

francocorreasosa: ajajajaja why

Issue #711, opened 2018-07-19
Dialog fixes


landitus:
The Cosmos dialog is slightly different that the one from the product. Here are the main differences to make it look alike:

- [ ] The dialog should align to the top of the page. We can use the closest value from our spacing tokens. For an example, check: Create Users
- [ ] Check for a very long dialog and how scrolling is achieved. For an example check: https://manage.auth0.com/#/connections/social and then Facebook.
- [ ] Add an example of dialog long form (like https://manage.auth0.com/#/connections/social) in Storybook
- [ ] Check the background color that opaques the page (we could set a default but leave it as a prop)

<img width="1082" alt="screen shot 2018-07-19 at 17 40 33" src="https://user-images.githubusercontent.com/239215/42968885-fb37891c-8b7a-11e8-944b-13aa418b98f4.png">

All of this can be tested in the example app as well.

PR #1102, opened 2018-11-20
Dialog component

andresgalante:
This PR makes the dialog component responsive to the size of the viewport, allowing scrolling when the content is larger than the screen.

It also makes the dialog accessible

And adds a new btn appearance called "action" that makes a btn look like regular text, mostly to be used in icons.

Closes: https://github.com/auth0/cosmos/issues/711


core/components/atoms/_overlay/overlay.js
```diff
@@ -13,6 +14,12 @@ const keyCodes = {
   escape: 27
 }
 
+export const overlayContentSizes = {
```
siddharthkp: // TODO for laters: tokenise these

core/components/atoms/_overlay/overlay.js
```diff
@@ -21,6 +28,12 @@ class Overlay extends React.Component {
     this.contentElement = null
   }
 
+  static getSizeForOverlay(propValue) {
+    if (typeof propValue === 'number') return `${propValue}px`
```
siddharthkp: Do we want to support an open API for this?

core/components/atoms/_overlay/overlay.js
```diff
@@ -76,22 +91,28 @@ Overlay.Backdrop = styled.div`
   bottom: 0;
   right: 0;
   z-index: ${layers.overlayBackdrop};
-  background: rgba(0, 0, 0, 0.7);
+  background: hsla(0,12%,95%,.95);
```
siddharthkp: rgba -> hsla? 

I'm a color noob, should we prefer one over the other?

core/components/molecules/dialog/dialog.js
```diff
@@ -34,75 +35,165 @@ const createButtonForAction = (action, index) => {
...
-      <DialogBody {...Automation('dialog.body')}>{props.children}</DialogBody>
-      <DialogFooter {...Automation('dialog.footer')}>
-        <ButtonGroup>{props.actions.map(createButtonForAction)}</ButtonGroup>
-      </DialogFooter>
-    </DialogElement>
-  </Overlay>
-)
-
-const DialogElement = styled.div`
+const focusOnFormInput = ({ current }) => {
```
siddharthkp: This is pretty form + dialog specific code. Can this be done with an `autoFocus` prop by the user instead?

core/components/molecules/dialog/dialog.js
```diff
@@ -34,75 +35,165 @@ const createButtonForAction = (action, index) => {
...
+  const form = node.querySelector('form')
+  if (!form) return
+
+  const firstInput = form.querySelector('input')
+  if (!firstInput) return
+
+  firstInput.focus()
+}
+
+const roleDependantProp = (props, requiredRole, propObject) =>
```
siddharthkp: Sorry, what does this do?

I'm also an accessibility noob 😅

core/components/molecules/dialog/dialog.js
francocorreasosa: This is the idea, for us to force this to play well with the other dialog-specific accessibility enhancements.

core/components/molecules/dialog/dialog.js
```diff
@@ -34,75 +35,165 @@ const createButtonForAction = (action, index) => {
...
   position: relative;
-  overflow: hidden;
-  padding: ${spacing.small};
-  padding-bottom: 0;
+  padding: ${spacing.small} ${spacing.large} ${spacing.xsmall} ${spacing.large};
   color: ${colors.text.default};
-  font-weight: ${fonts.weight.medium};
+  word-break: break-word;
+  white-space: pre-wrap;
+  word-wrap: break-word;
```
siddharthkp: I'm trusting @andresgalante to know what's going on here 😋

core/components/molecules/dialog/dialog.js
francocorreasosa: Basically sets a prop depending on the dialog's role (See auth0/cosmos/pull/1102/files/4ea9b610e8a523e003c6c7f757289d30777eb4fe#diff-dedf8280fcbfc0d57cb9fbfb4df7f85cR72)

core/components/molecules/dialog/dialog.js
```diff
@@ -34,75 +35,165 @@ const createButtonForAction = (action, index) => {
...
-  }
-  & ${Icon.Element} {
-    flex: none;
-    cursor: pointer;
-  }
+`
+
+const DialogTitle = styled(Heading)`
+  font-weight: ${fonts.weight.medium};
+  font-size: 1em;
```
siddharthkp: 1. Do we have a system wide preference of `px` vs `em`?

2. Should components have a value either way or should this be picked from the tokens: example: `tokens.fonts.size.default`

core/components/molecules/dialog/dialog.js
```diff
@@ -34,75 +35,165 @@ const createButtonForAction = (action, index) => {
...
 Dialog.propTypes = {
   actions: PropTypes.arrayOf(
     PropTypes.oneOfType([PropTypes.instanceOf(DialogAction), PropTypes.element])
-  ).isRequired,
-  title: PropTypes.string.isRequired,
-  width: PropTypes.number,
-  onClose: PropTypes.func
+  ),
+  title: PropTypes.string,
+  titleElement: PropTypes.oneOf([1, 2, 3, 4]),
```
siddharthkp: We can document the prop types with comments

Example: https://github.com/auth0/cosmos/blob/master/core/components/atoms/button/button.js#L292-L322

core/components/molecules/dialog/dialog.md
```diff
@@ -82,3 +81,38 @@ class DialogContainer extends React.Component {
   }
 }
 ```
+
+## Usage
```
siddharthkp: This is beautiful ✨

core/components/molecules/dialog/dialog.story.js
```diff
@@ -13,10 +13,11 @@ storiesOf('Dialog').add('default', () => (
     <Dialog
       open
       title="Example Dialog"
-      onClose={() => {}}
+      role="destructive"
+      onClose={() => { }}
       actions={[
-        new Dialog.Action('OK', () => {}, 'primary'),
-        new Dialog.Action('Cancel', () => {})
+        new Dialog.Action('OK', () => { }, 'primary'),
```
siddharthkp: Not for this PR, but @francocorreasosa Do you think we should replace this with just buttons™

francocorreasosa: Of course 😉

core/components/molecules/dialog/dialog.js
siddharthkp: Ah got it, can we call the function `getAccessibilityRole`?

francocorreasosa: If you want.. That would not be a problem for me

core/components/atoms/_overlay/overlay.js
andresgalante: I don't think it's a big deal, this is to match exactly the color we have today at style guides.

We'll standarize once we create the variable system, hopefully soon enough

core/components/molecules/dialog/dialog.js
andresgalante: This is saying: if the word is a very very very long string without spaces, then break it and wrap it into a new line. You can check the stressed example.

core/components/molecules/dialog/dialog.js
andresgalante: Yes, good catch. This will be easier when we do have a variable system in place.

We'll also standarize in using `rem` with a px to rem function. something like this: https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/sass-utilities/functions.scss

core/components/molecules/dialog/dialog.md
```diff
@@ -82,3 +87,38 @@ class DialogContainer extends React.Component {
...
+Try to avoid dialogs for:
+
+- **Passive:** Dialogs that can be replaced for notifications.
+- **Input:** Dialogs that include forms.
+
+When you see the need to create a dialog, ask yourself if this interaction must completly stop the user's flow. If the answer is no, then consider other options to display the information.
+
+## Examples
+
+- Default passive modal (missing code snippet)
```
siddharthkp: Is this pending or should we remove this from the docs?

core/components/molecules/dialog/dialog.story.js
```diff
@@ -39,14 +46,21 @@ storiesOf('Dialog').add('with primary button only', () => (
 
 storiesOf('Dialog').add('with secondary button only', () => (
```
siddharthkp: The story is called `with secondary button only` but has 2 buttons. Create a new story please

core/components/molecules/dialog/dialog.story.js
```diff
@@ -39,14 +46,21 @@ storiesOf('Dialog').add('with primary button only', () => (
...
   <StyledExample title="with secondary button only">
+    <Button>test</Button>
```
siddharthkp: Leftover debug statement?

core/components/molecules/dialog/dialog.story.js
```diff
@@ -39,14 +46,21 @@ storiesOf('Dialog').add('with primary button only', () => (
...
+      actions={[
+        <Button onClick={() => {}}>No</Button>,
+        <Button onClick={() => {}} appearance="destructive">
+          Yes, continue
+        </Button>
+      ]}
     >
       Are you sure?
     </Dialog>
+    <Button>test</Button>
```
siddharthkp: Are these intentional?

core/components/molecules/dialog/dialog.js
```diff
@@ -34,75 +35,174 @@ const createButtonForAction = (action, index) => {
...
-  width: PropTypes.number,
-  onClose: PropTypes.func
+  ),
+  /** Dialog's header title */
+  title: PropTypes.string,
+  /** Dialog's header title heading element */
+  titleElement: PropTypes.oneOf([1, 2, 3, 4]),
+  /** Dialog's container width */
+  width: PropTypes.oneOfType([PropTypes.number, PropTypes.oneOf(Object.keys(overlayContentSizes))]),
+  /** Called when the user clicks on the header's cross or outside the dialog */
```
siddharthkp: I guess this definition is not fully accurate because the dialog can also be closed from the keyboard.

Let's just rephrase this to: "Callback triggered when the the dialog is closed by the user"

core/components/molecules/dialog/dialog.js
```diff
@@ -34,75 +35,174 @@ const createButtonForAction = (action, index) => {
...
     PropTypes.oneOfType([PropTypes.instanceOf(DialogAction), PropTypes.element])
-  ).isRequired,
-  title: PropTypes.string.isRequired,
-  width: PropTypes.number,
-  onClose: PropTypes.func
+  ),
+  /** Dialog's header title */
+  title: PropTypes.string,
+  /** Dialog's header title heading element */
+  titleElement: PropTypes.oneOf([1, 2, 3, 4]),
```
siddharthkp: Needs a better name.

Looking at the prop name, I didn't realise this is what <h*> tag will be used. I'd recommend using a prop that matches the underlying `Heading` component: https://auth0-cosmos.now.sh/docs/#/component/heading

```js
/** Title in header */
title: PropTypes.string,
/** size of `Heading` used in title */
titleSize: PropTypes.oneOf([1, 2, 3, 4])
```

`headingSize` could also work but it wouldn't feel related to `title` then

core/components/molecules/dialog/dialog.js
```diff
@@ -34,75 +35,174 @@ const createButtonForAction = (action, index) => {
...
+  const form = node.querySelector('form')
+  if (!form) return
+
+  const firstInput = form.querySelector('input')
+  if (!firstInput) return
+
+  firstInput.focus()
+}
+
+const getAccessibilityRole = (props, requiredRole, propObject) =>
```
siddharthkp: Please expand this function to be more easier to understand and add comments for what it does

core/components/molecules/dialog/dialog.js
```diff
@@ -34,75 +35,174 @@ const createButtonForAction = (action, index) => {
...
+    const props = this.props
+    return (
+      <Overlay contentSize={props.width} {...props}>
+        <FocusTrap persistentFocus={false} onExit={props.onClose}>
+          <DialogBox
+            width={props.width}
+            {...Automation('dialog')}
+            {...getAccessibilityRole(props, 'destructive', {
+              'aria-describedby': 'dialog-description'
+            })}
```
siddharthkp: 😲🤷‍♀️

see comment on `getAccessibilityRole` above

core/components/molecules/dialog/dialog.js
```diff
@@ -34,75 +35,174 @@ const createButtonForAction = (action, index) => {
...
+  /** Dialog's header title */
+  title: PropTypes.string,
+  /** Dialog's header title heading element */
+  titleElement: PropTypes.oneOf([1, 2, 3, 4]),
+  /** Dialog's container width */
+  width: PropTypes.oneOfType([PropTypes.number, PropTypes.oneOf(Object.keys(overlayContentSizes))]),
+  /** Called when the user clicks on the header's cross or outside the dialog */
+  onClose: PropTypes.func,
+  /** Whether you're presenting a form or a destructive action */
+  role: PropTypes.oneOf(['form', 'destructive'])
```
siddharthkp: Is there a custom role possible for other dialog cases?

Also, please add examples to explain what this means and does. I couldn't figure what this is

core/components/molecules/dialog/dialog.js
siddharthkp: Sorry, I don't understand what you mean. Can you add a comment in the code explaining it?

core/components/molecules/dialog/dialog.story.js
francocorreasosa: Intentional, to test focus trap

core/components/molecules/dialog/dialog.story.js
francocorreasosa: Intentional, to test focus trap

core/components/molecules/dialog/dialog.js
francocorreasosa:
```suggestion
  /** Callback triggered when the the dialog is closed by the user */
```

core/components/molecules/dialog/dialog.js
francocorreasosa: The ultimate goal is not to relate the heading element to the size. So we went with this approach to be covered if in the future we make the heading element independent of the size, this is one less prop to deprecate.

core/components/molecules/dialog/dialog.md
francocorreasosa: This is the default role of the dialog (general content). I will add the missing snippet.

core/components/molecules/dialog/dialog.js
francocorreasosa: The role of the dialog helps to define where the focus should be at the time of presentation or to add additional accesibility tags. You need to set it to `form` when you have a form inside so the focus will go to the first input. In case is `destructive` the focus will be on the close button (as default) but the `aria-describedby` attribute will be set to `dialog-description`.

On other cases, no role is required. You can use the dialog as normal.

francocorreasosa: Stories for both roles and default are present.

core/components/molecules/dialog/dialog.js
francocorreasosa: Done

core/components/molecules/dialog/dialog.story.js
francocorreasosa: Done

core/components/atoms/_overlay/overlay.js
francocorreasosa: Not for now

core/components/atoms/_overlay/overlay.js
francocorreasosa: Agree, we're going to wait till we have the 2nd version of tokens.

core/components/molecules/dialog/dialog.md
landitus: Checking on the props for `role` there are only two options: form and destructive. I'm guessing we are missing a `general` or `content` (even as a default) prop for the use cases where none of these two apply. Still, the user may not understand why this role needs to be set.

francocorreasosa: Added `default` role as default.

core/components/molecules/dialog/dialog.js
siddharthkp: Hi :wave: 

Saw this was resolved, but couldn't find the comment

core/components/molecules/dialog/dialog.js
siddharthkp: comment should explain *why* not *what* 😢

core/components/molecules/dialog/dialog.js
siddharthkp: Got it! This isn't related to size at all, but only the h tag.

To make this obvious, can we change the accepted values to 

```js
titleElement: PropTypes.oneOf(['h1', 'h2', 'h3', 'h4'])
```

siddharthkp: Moving this conversation to the conversation tab

siddharthkp:
API decisions in this PR:

1. The usage of `titleElement` is not super obvious. 

    After reading the description and available props, I expected it to change the size of the Heading.

image


    That was wrong and later explained in the docs under `Accessibility` - Make sure to adjust the heading level according to what makes sense on your document tree.


    To make this obvious, 

	1. Can we change the accepted values to

	```js
	titleElement: PropTypes.oneOf(['h1', 'h2', 'h3', 'h4'])
	```

	(so that folks don't tie this with size prop of Heading)

   2. What can add we add to the description? Will adding _for accessibility_ help?

	```
	titleElement: Dialog's header title heading element (for accessibility)
	```

2. What should be the default `role`?

   There are 2 use case specific roles: ['form', 'destructive']. This is a choice the user has to make.

   If they don't make that choice, we need to make a choice for them. Right now, it's `default` - which doesn't really do anything and feels like folks won't bother changing it (we don't have a lot of accessibility heros)

	1. Should we change `default` to `content` which is more expressive (and make it the default)?

	2. Is `content` even a valid role? If you are using Dialog to show content, should you even be using a dialog? cc: @andresgalante

andresgalante:
@siddharthkp 
1-  Agreed! and I wouldn't put "for accessibility" between ( ).

```
titleElement: Dialog's header title heading element for accessibility
```

and

2- Although no one should use the dialog for content, but it is been used for content.

The specs have a very specific definition for destructive vs content, but it is subtle and they can be the same. I agree, we should have a `default` option.

@francocorreasosa I am sorry I am in meetings all day during this week, can you make these changes plase? or @siddharthkp can you patch it?

siddharthkp:
> The specs have a very specific definition for destructive vs content, but it is subtle and they can be the same. I agree, we should have a default option.

Should we call the 3rd role - `content` or `default`?

andresgalante:
I'd call it `default`. the specs since it's everything that's not destructive or form.

core/components/molecules/dialog/dialog.js
francocorreasosa: np, added the 'why' as well :)

landitus:
> Can we change the accepted values to
titleElement: PropTypes.oneOf(['h1', 'h2', 'h3', 'h4'])

Indeed, we did this change today with @francocorreasosa . Now, it's more obvious that you are choosing the element to be rendered. Good catch @siddharthkp 	!!

PR #1104, opened 2018-11-20
Fix full logo color and complete the docs for the new variant

francocorreasosa:


landitus:
@francocorreasosa I've realized this PR and last PR will create a bunch of problems with the Logo when we added the full version. Mostly it's because there are some combinations we shouldn't allow, and the props create some confusion: full/default, the color options as well. 

I want to do a better spec first so that we can get the design sorted out. I'm sorry for this as I'm realizing now! 

Can we revert https://github.com/auth0/cosmos/pull/1097?

francocorreasosa:
Yep, no problem. Just merge #1118 and it will be reverted.

francocorreasosa:
Closing this in the meantime.

PR #1135, opened 2018-11-23
 Fix applications details warning to be thrown by a missing prop via @jvelezpo

francocorreasosa:


francocorreasosa:
Thank you @jvelezpo!

Issue #1131, opened 2018-11-22
Typescript support/definitions not working


landitus:
**Describe the bug**
I've heard from @hzalaz that the current implementation to support typescript is failing. @jcenturion gave us a head start, but we need to either own it or drop it.

**Expected behavior**
Let's check with our users, test it out and find a solution.

**Cosmos version:**
0.8.1

francocorreasosa:
I'm actually working on this

landitus:
Awesome!

PR #1136, opened 2018-11-23
Remove prop-types to typescript declarations module

francocorreasosa:
This is a temporal _solution_. 

I want to make a manual typing proof of concept anytime soon, which involves adding a TypeScript compilation pipeline to Cosmos.

Related to #1131 (but does not resolves it)

### DIFF
--- a/core/components/package.json
+++ b/core/components/package.json
@@ -8,7 +8,6 @@
   "keywords": [],
   "author": "auth0",
   "license": "MIT",
-  "types": "meta/index.d.ts",
   "dependencies": {
     "md5": "^2.2.1",
     "@auth0/cosmos-tokens": "0.8.1",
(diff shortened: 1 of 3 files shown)


landitus:
Sounds good @francocorreasosa 	!!

siddharthkp:
👍 

Is this step 1 of solving the issue? 

Step 1 - Removing buggy implementation
Step 2 - Create working implementation

francocorreasosa:
I'm not completely sure, IMO we should not merge this till we have the other solution. But if this issue is difficulting app development, we should merge it to get out of their way. WDYT? @jcenturion 	@cristiandouce 	@hzalaz

hzalaz:
@francocorreasosa its ok, we are overriding it ourselves to avoid the error so we could remove that from our code too

Issue #1082, opened 2018-11-13
Add support for prepend and append content to Text Input


landitus:
We need the ability to enhance the **text input** with units, icons and text. 

**Examples:**
_____ .auth0.com
_____ seconds
Icon _____
$ ________ (this example doesn't exist but it wouldn't hurt to add it)
 
<img width="613" alt="screen shot 2018-11-13 at 13 01 53" src="https://user-images.githubusercontent.com/239215/48415140-79ad6800-e744-11e8-9558-f6326b64ba24.png">

![image](https://user-images.githubusercontent.com/1863771/49011750-471c5a00-f19d-11e8-8b72-7147dc77c32a.png)

PR #1142, opened 2018-11-26
add a tiny story

siddharthkp:
### Problem

`Form` fields have a pretty confusing API

image

Playground link for the above code: [here](https://auth0-cosmos.now.sh/docs/#/playground?code=const%20Example%20=%20()%20=%3E%20%7B%250A%20%20return%20(%250A%20%20%20%20%3CForm%3E%250A%20%20%20%20%20%20%3CForm.TextInput%250A%20%20%20%20%20%20%20%20label=%22Allowed%20URLs%22%250A%20%20%20%20%20%20%20%20type=%22text%22%250A%20%20%20%20%20%20%20%20placeholder=%22Enter%20something%22%250A%20%20%20%20%20%20%20%20defaultValue=%22auth0.com%22%250A%20%20%20%20%20%20%20%20error=%22This%20is%20not%20a%20valid%20URL%22%250A%20%20%20%20%20%20%20%20helpText=%22Make%20sure%20to%20specify%20the%20protocol,%20http://%20or%20https://%22%250A%20%20%20%20%20%20%20%20actions=%7B%5B%250A%20%20%20%20%20%20%20%20%20%20%7B%20icon:%20%22copy%22,%20label:%20%22Copy%20URL%22,%20handler:%20e%20=%3E%20console.log(e)%20%7D%250A%20%20%20%20%20%20%20%20%5D%7D%250A%20%20%20%20%20%20/%3E%250A%20%20%20%20%3C/Form%3E%250A%20%20)%250A%7D%250A%250Arender(Example)%250A)

It looks deceptively similar to `TextInput` API, which was intentional in the original API design, but after hearing from our users (developers), I realised this leads to more confusion than simplicity.

Example:

- Which of these props are for the Form Field and which ones are passed to `TextInput`?

	Few take out a few props for `Form.Field` and pass all the others to the field component inside.

    In the above example, `label`, `helpText` are Field props while `type`, `placeholder` and `defaultValue` are TextInput props.

- Do I pass `error` or `hasError`? 

    `TextInput` accepts a `hasError` boolean while `Form.TextInput` accepts `error` as string. This is intentional so that the Field can show the error message and pass `hasError` to the `TextInput` implicitly. 

    But, It's easy to pass `error` to `TextInput` and expect it to add an error message because they API looks similar

- `actions` is a Field prop but only applies to `Form.TextInput` (not `Form.TextArea`, `Form.Switch` etc.)

    This is going to become even more confusing soon when `TextInput` outside Form starts accepting actions (https://github.com/auth0/cosmos/issues/1082)

### Proposal

Make Form field more explicit to differentiate

### DIFF
--- a/core/components/molecules/form/field/stories/text-field.story.js
+++ b/core/components/molecules/form/field/stories/text-field.story.js
@@ -7,7 +7,9 @@ import { Form } from '@auth0/cosmos'
 storiesOf('Form').add('text field', () => (
   <Example title="text field">
     <Form>
-      <Form.TextInput label="Field label" type="text" placeholder="Enter something" />
+      <Form.Field label="Field label">
+        <TextInput type="text" placeholder="Enter something" />
+      </Form.Field>
     </Form>
   </Example>
 ))


diptigandhi24:
> @diptigandhi24 What do you think?

Yes that works!! but I still have to keep track of props that can be passed to formFields and FieldComponent, right?

siddharthkp:
> Yes that works!! but I still have to keep track of props that can be passed to formFields and FieldComponent, right?

There are some (3) props that are for Field wrapper for label, helpText and error.

But other than that, you can use any field like `TextInput` with their native props.

diptigandhi24:
> > Yes that works!! but I still have to keep track of props that can be passed to formFields and FieldComponent, right?
> 
> There are some (3) props that are for Field wrapper for label, helpText and error.
> 
> But other than that, you can use any field like `TextInput` with their native props.

oh yeah makes sense to have those props for field wrapper, I see the point this should definitely work!!

nicogarcia:
Hi! I agree with your stated Pros and Cons and I think Pros are more beneficial than cons perjudicial, having an abstraction over the fields seems like a good idea but it may be confusing and miseducate about how the structure of the Form components really is (Field > Text/Switch/whatever). So, I think your proposal is better than the old API!

nicogarcia:
Hi! I agree with your stated Pros and Cons and I think Pros are more beneficial than cons perjudicial, having an abstraction over the fields seems like a good idea but it may be confusing and miseducate about how the structure of the Form components really is (Field > Text/Switch/whatever). So, I think your proposal is better than the old API!

PR #1143, opened 2018-11-26
Use appearance none in select and manually render the icon

### LINKED ISSUE #1139: Select Component looks weird on Firefox
**Describe the bug**
Select component looks weird in Firefox vs Chrome.

**Expected behavior**
They should look more likely to each other.

**Screenshots**

image
☝️**Firefox**

image

☝️ **Chrome**

**Cosmos version:**
`0.8.1`

**Additional context**
Even weirder with custom impl in manage
image

francocorreasosa:
Resolves #1139

### DIFF
--- a/core/components/atoms/select/select.js
+++ b/core/components/atoms/select/select.js
@@ -1,7 +1,8 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import Automation from '../../_helpers/automation-attribute'
-import { css } from 'styled-components'
+import Icon from '../icon'
+import styled from 'styled-components'
 
 import { misc, colors } from '@auth0/cosmos-tokens'
 import { StyledInput } from '../_styled-input'
@@ -12,7 +13,17 @@ const selectOpacity = {
 }
 const PLACEHOLDER_VALUE = '__select_placeholder'
 
+const StyledIcon = styled(Icon).attrs({
+  name: 'dropdown-fill',
+  color: 'black',
+  size: '10'
+})`
+  position: absolute;
+  right: 0.8em;
+`
+
 const StyledSelect = StyledInput.withComponent('select').extend`
+  appearance:none;
   height: ${misc.input.default.height};
   opacity: ${props => (props.disabled ? selectOpacity.disabled : selectOpacity.default)};
   background-color: ${props =>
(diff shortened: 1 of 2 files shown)


landitus:
Cool, from the design perspective, this change will make the selects more consistent with the rest of the inputs. Adding a quick design reference to scale the icon and position it properly. Let me know if there's anything missing.

From the accessibility perspective, let's check the tips @andresgalante 	was mentioning.

screen shot 2018-11-27 at 10 57 23

Make sure the arrow icon uses the text color primary token :)

landitus:
@francocorreasosa 	Check this one: https://www.chromaticqa.com/snapshot?appId=5aba7f6a07f95800202658ff&id=5bfd55a863f5530024a45448

Looks like a rogue arrow to me! :)

landitus:
The arrow and position is great! But in the stressed example we need the elipsis to kick in. Didn’t we add it?

landitus:
Given the native Select doesn't have ellipsis, let's move forward as we have it :)

core/components/atoms/select/select.js
```diff
@@ -12,6 +13,22 @@ const selectOpacity = {
 }
 const PLACEHOLDER_VALUE = '__select_placeholder'
 
+const StyledIcon = styled(Icon).attrs({
+  name: 'dropdown-fill',
+  size: '14'
+})`
```
siddharthkp: Instead of extending the Icon component, can we use this as props in the place where it is used?

```jsx
<Icon name="dropdown-fill" size="14" />
```

and then add styles in the wrapper?

```js
Select.Wrapper = styled.div`
  ${Icon} {
    right: 12px;
  }
`
```

Whenever possible, we should prefer React API over styled-components API

### Why?

- Single consistent way of using components
- Makes it easier to upgrade (or replace) styled-components
- Set the right example for how we want folks to use cosmos components by using them right inside the system as well

core/components/atoms/select/select.js
```diff
@@ -12,6 +13,22 @@ const selectOpacity = {
...
+  position: absolute;
+  right: 12px;
+
+  svg {
+    display: block;
+
+    path {
+      fill: ${colors.text.default};
+    }
```
siddharthkp: Looks like this could be a `color` prop as well?

francocorreasosa: Yes!

core/components/atoms/select/select.js
francocorreasosa: Completely agree with you, just fixed it :)

francocorreasosa:
@siddharthkp Can you review again when you have a chance?

Issue #1081, opened 2018-11-13
Add missing icons


landitus:
Review which icons are being used right now and add them in bulk to the Icon library.

francocorreasosa:
The following (from Manage) are missing:

- office365
- facebook
- google-oauth2
- google-apps
- windowslive
- yahoo
- aol
- linkedin
- paypal
- paypal-sandbox
- dropbox
- github
- bitbucket
- twitter
- amazon
- vkontakte
- yandex
- thirtysevensignals
- box
- salesforce
- salesforce-sandbox
- salesforce-community
- fitbit
- baidu
- renren
- weibo
- shopify
- dwolla
- miicard
- wordpress
- yammer
- soundcloud
- instagram
- bitly
- evernote
- evernote-sandbox
- flickr
- thecity
- thecity-sandbox
- planningcenter
- exact
- daccount
- sharepoint
- ip
- adfs
- pingfederate
- waad
- auth0-adldap
- ad
- custom
- auth0
- samlp
- sms
- email

PR #1146, opened 2018-11-27
Add auth0 social icons (2nd round)

francocorreasosa:
Closes: https://github.com/auth0/cosmos/issues/1081


landitus:
This LGTM. @francocorreasosa 	I opted to not add more aliases as these belong to the application/product. So we are good to merge

PR #1161, opened 2018-11-28
[Table] Add accessibility to table header and table row by @rfbotto

### LINKED ISSUE #872: Table component is not keyboard accessible

siddharthkp:
Copied changes from #1057 because it was stale

Fixes #872

### DIFF
--- a/core/components/molecules/table/table-header.js
+++ b/core/components/molecules/table/table-header.js
@@ -18,7 +18,7 @@ const TableHeader = props => {
       )
     }
 
-    const onClick = column => {
+    const sortColumn = column => {
       if (!column.sortable) return
 
       let sortDirection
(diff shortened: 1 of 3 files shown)


siddharthkp:
Accessibility question:

This PR adds `tabIndex={0}` to every row :thinking_face:

1. do html tables have some accessibility baked in?
2. how do we decide if hitting tab should go to the next row or next cell?
3. will we do more harm than good if we add tabIndex ourselves?
4. should this work with arrow keys instead? excel style (edited)
5. we probably should implement Enter => row click as well

andresgalante:
Take a look at this PR later, when we create a fully accessible table

PR #1162, opened 2018-11-28
Ignore storybook cache

francocorreasosa:


andresgalante:
Lets merge it as soon as chromatic is happy

Issue #1166, opened 2018-11-28
Docs components are missing the line-height


landitus:
**Describe the bug**
After we removed the line-height from the global styles, we now have some visual regression on components that live in the Documentation. From what I can see, it's the Playground and the Sidebar.

**Expected behavior**
The Playground and the Sidebar should use the global line-height value.

**Screenshots**
Check below

**Cosmos version:**
master

<img width="1274" alt="screen shot 2018-11-28 at 20 42 07" src="https://user-images.githubusercontent.com/239215/49189796-7de5a200-f34e-11e8-9c46-8e25537cf1f6.png">

PR #1167, opened 2018-11-28
Applied container styles to Sidebar, Codeblock and Playground

landitus:
Addresses: https://github.com/auth0/cosmos/issues/1166

I've tried added the container styles to the affected components, but I'm not sure if this is the answer for this situation. Should we just declare the line-height here, without the helper?

@siddharthkp 	can you check it out?

### DIFF
--- a/internal/docs/docs-components/code-block.js
+++ b/internal/docs/docs-components/code-block.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import styled from 'styled-components'
-import containerStyles from '../../../core/components/_helpers/container-styles'
+import containerStyles from '@auth0/cosmos/_helpers/container-styles'
 
 import Highlight from 'react-highlight'
 import './material-theme.css'
--- a/internal/docs/docs-components/code-block.js
+++ b/internal/docs/docs-components/code-block.js
@@ -1,10 +1,12 @@
 import React from 'react'
 import styled from 'styled-components'
+import containerStyles from '../../../core/components/_helpers/container-styles'
 
 import Highlight from 'react-highlight'
 import './material-theme.css'
 
 const Wrapper = styled.div`
+  ${containerStyles};
   margin: 16px 0;
   code {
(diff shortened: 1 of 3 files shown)


internal/docs/docs-components/code-block.js
```diff
@@ -1,10 +1,12 @@
 import React from 'react'
 import styled from 'styled-components'
+import containerStyles from '../../../core/components/_helpers/container-styles'
```
siddharthkp: nuuuuu, this breaks whenever make the tiniest change to folder structure

```suggestion
import containerStyles from '@auth0/cosmos/_helpers/container-styles'
```

This way, it's more reliable

siddharthkp: nuuuuu, this breaks whenever make the tiniest change to folder structure

```suggestion
import containerStyles from '@auth0/cosmos/_helpers/container-styles'
```

This way, it's more reliable

internal/docs/sidebar/index.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react'
 import styled from 'styled-components'
+import containerStyles from '../../../core/components/_helpers/container-styles'
```
siddharthkp: Explained in the other comment

```suggestion
import containerStyles from '@auth0/cosmos/_helpers/container-styles'
```

internal/docs/spec/playground.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react'
 import styled from 'styled-components'
+import containerStyles from '../../../core/components/_helpers/container-styles'
```
siddharthkp: Explained in the other comment

```suggestion
import containerStyles from '@auth0/cosmos/_helpers/container-styles'
```

siddharthkp:
> Should we just declare the line-height here, without the helper?

I like using the helper, cosmos docs built with cosmos gives me joy 😄

PR #1171, opened 2018-11-29
Add label position for switch by @m4tr1k

### LINKED ISSUE #1083: Support for text position change in Switch

siddharthkp:
Fixes #1083


siddharthkp:
@andresgalante Great PR! The alignment fix only works when the switch is on it's own. It seems to misalign things when used with an error message - Check out the denied chromatic changes

andresgalante:
@siddharthkp yeah... the space between a switch and the error changes.

The way I see it it's a problem on the form comoponent. I don't know if we should embarque in a refactoring of form attached to this PR.

Do you think we can merge the swicth as is, and scechule a revision of the form component's HTML and CSS for later? not just a clean up but also to make it responsive and accecible.

siddharthkp:
> Do you think we can merge the swicth as is, and scechule a revision of the form component's HTML and CSS for later? not just a clean up but also to make it responsive and accecible.

Nope, if there is a bug in the form that was hidden until now, it will surface through this PR.

I'd say we should fix that first, then merge this. Leaving master in a broken state is risky because we might have to do a bugfix release in between.

> I don't know if we should embarque in a refactoring of form attached to this PR.

I wouldn't recommend a full refactor with responsive + accessible, just enough to hide this bug from surfacing

siddharthkp:
> @siddharthkp Can you resolve the conflicts, maybe something is out of date?

Doesn't exist on master.

andresgalante:
@landitus @siddharthkp the last push I sent fixes the vertical spacing.

PR #1172, opened 2018-11-29
Button: Add type prop + Use it in secondary Form.Actions

### LINKED ISSUE #1163: Form.Actions - Allow passing button props
**Describe the bug**
In order to better control the submit flow we need to be able to fully customize the type of buttons used in forms. As it's recommended from cosmos to use `Form.Actions` it'd ideal that:

1. We could customize the `type` attribute of the actions
2. That secondary actions (buttons) have type "button" by default.

**Expected behavior**
When the button is type `submit` the control of the form submission passes through it's state. So, if for example the button is disabled, the whole form is disabled from submission.

https://codepen.io/anon/pen/MzPGKj

**Screenshots**
Check the CodePen above.

**Cosmos version:**
`0.8.1`

**Additional context**
It's crucial for application flow control.

siddharthkp:
Fixes https://github.com/auth0/cosmos/issues/1163

```diff
Button.propTypes = {
+ type: PropTypes.oneOf(['button', 'submit', 'reset'])
}

Button.defaultProps = {
+ type: 'submit'
}
```

The default matches [the HTML spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#Attributes), but should we add a default at all? Is it a browser implementation detail that we should let the browser take care of?

@francocorreasosa @andresgalante what do you think?

-- 

To fix #1163, secondary actions in Form use `type=button` now.

--

Unrelated change: Deleted generated docs file, this should not have been on github


siddharthkp:
@andresgalante / @francocorreasosa Open question here:

The browser attaches a default (submit) on it's own. Should we leave it to the browser to take care of and not add any default in cosmos?

PR #1192, opened 2018-11-30
Fixes Resource List columns

### LINKED ISSUE #523: Fix columns on ResourceList
Have some feedback on the new ResourceList component:

- Improve handling of large subtitle. Currently there is a bug that causes this:
image

Ideally, it would look like this:
image

- Row or Title click handlers...  would be very helpful to allow making the row or title clickable, in addition to the actions. Ex, clicking the row could take you to a child page with the item's details, while the quick actions are for immediate actions...

andresgalante:
This PR fixes #523 by cleaning up the resource list CSS.

It allows the first column to extend when there is no second column :)

### DIFF
--- a/core/components/atoms/code/code.js
+++ b/core/components/atoms/code/code.js
@@ -14,6 +14,7 @@ Code.Element = styled.span`
   background-color: ${colors.base.blueLightest};
   border-radius: 3px;
   line-height: 18px;
+  word-break: break-all;
 `
 
 Code.propTypes = {}
--- a/core/components/molecules/resource-list/item/item.js
+++ b/core/components/molecules/resource-list/item/item.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import styled from '@auth0/cosmos/styled'
+import styled from 'styled-components'
 import { Button, ButtonGroup, Link } from '@auth0/cosmos'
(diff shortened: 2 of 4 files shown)


andresgalante:
This PR has been updated. The component now can handle any amount of columns.

The problem we are seen is that it'll not align the middle column if the amount of action is not consistent. After talking with @landitus a solution can be to force always the same amount of actions but have them disabled if necessary. Will see if that pans out.

andresgalante:
I think so, if we merge it as is, then it would cause some problems to people using the resource list, maybe we can spend a few more days on it and fix it? 

would you rather merge this one and open a new one for the columns problem?

landitus:
As @andresgalante 	says, I believe we should not merge it yet if we foresee issues. I'm going to do a design review and report back to check if we merge.

landitus:
FYI, will browsing the docs, I get this error and the example is not being rendered:

`ReferenceError: callHandler is not defined`

screen shot 2018-12-10 at 20 14 55

andresgalante:
It got solved after rebasing.

I am waiting for the sizes to continue with this PR, so far I got `1/2 1/2` `1/3 2/3` and `2/3 1/3` for `distribution` between header and body.

Actions are always optional.

andresgalante:
@landitus I've removed the distribution prop and it should have all the things we've talked about

PR #1194, opened 2018-12-01
Do not show `children` props table

### LINKED ISSUE #676: Props table in docs shouldn't show children as a prop
Right now, we aren't handling the `children` prop that well in docs. For example, in `Heading`:

screen shot 2018-07-09 at 11 43 56 am

This same behavior is shown in `ButtonGroup`. This is a little confusing, and makes it so the code generated in the playground is incorrect if it's copied and pasted.

Does it make sense to assign `children` a prop type other than `Node` anyway?

(via @jcenturion)

tsriram:
Closes #676 

I'm not sure if this is the right way to do. I thought it's better not to have `children` prop in generated metadata than skipping the `children` prop in `Props` component's loop. Let me know if you want to handle this differently. Or feel free to close this if it's not required :)

### DIFF
--- a/internal/docs/spec/props.js
+++ b/internal/docs/spec/props.js
@@ -68,7 +68,7 @@ class Props extends React.Component {
 
     const keys = Object.keys(propData).filter(key => key[0] !== '_')
 
-    return (
+    return keys.length ? (
       <Table>
         <thead>
           <tr>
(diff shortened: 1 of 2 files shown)


tsriram:
Looks like the tests are failing because of bad credentials for GitHub API. Not sure if I could fix this :thinking:

siddharthkp:
Thanks!

Merging this into a local branch to take it forward

PR #1204, opened 2018-12-03
Docs: Do not show `children` in props table by @tsriram

### LINKED ISSUE #676: Props table in docs shouldn't show children as a prop

siddharthkp:
Closes #676

### DIFF
--- a/internal/docs/spec/props.js
+++ b/internal/docs/spec/props.js
@@ -68,7 +68,7 @@ class Props extends React.Component {
 
     const keys = Object.keys(propData).filter(key => key[0] !== '_')
 
-    return (
+    return keys.length ? (
       <Table>
         <thead>
           <tr>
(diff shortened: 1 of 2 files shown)


siddharthkp:
On second thoughts, we should not remove `children` from the table.

It's pretty useful in components that accept other components: https://auth0-cosmos.now.sh/docs/#/component/breadcrumb

Instead we should try to make it more useful: 


```js
  children: PropTypes.instanceOf(Breadcrumb.Link)
```

should be represented as 

```
children: React node of type Breadcrumb.Link
```

tsriram:
Makes sense :+1:

Issue #1105, opened 2018-11-21
RFC - overrides prop


siddharthkp:
#### This RFC (request for comments) is to introduce a way for applications to override styles of a cosmos component

&nbsp;

### Problem statement

Cosmos takes care of the _design_ of a component, ideally an application using cosmos should not modify the component's styles.

Sometimes, a new design is _slightly_ different from existing components.

Example: The team has to build a search input but it does not exist yet.

![image](https://user-images.githubusercontent.com/1863771/48810367-e74b2c80-ed20-11e8-92ef-0184d6829e7a.png)

Person A: This new component does not exist, I should start with an html `input` and build it locally.

Person B: This looks almost like `TextInput` from cosmos, I should take that and change some styles.

I don't blame Person B, sounds like an effective approach.

```jsx
const InputSearch = styled(TextInput)`
  background-color: #f6f6f6;
`;
```

(real code from auth0/manhattan)

This is a good example of overriding a cosmos component, here's another example:

```jsx
const TitleGroup = styled.div`
  display: flex;
  align-items: center;
  p {
    font-size: 13px
  }
`;
```

(roughly the same code from auth0/manhattan, modified for creating a relevant example)

The interesting thing from the above code sample is the `p` is coming from cosmos! This is also an override with a similar purpose. Cosmos does not support small text size in paragraph (yet?) and the design requires it.

&nbsp;

&nbsp;

### But, why do we allow overrides at all?

That's a great question.

If we could review all the design mockups and pull requests, there would be no need of overrides. But, with the speed and number of teams we aim to support, we will never be ahead of the teams in terms of components. 

However, we should be able to spot patterns and bring those into the system if needed.

There are also cases, where the team is building something on an old non-system design instead of a new design that uses cosmos (again, manhattan is the example)

We need to accept that overrides solve a purpose and they will exist. If we look at them periodically, we would be able to pick feature requests for cosmos or find bugs/inconsistency in the design.

_Note: This proposal is not for margins, see #1059 for that._

&nbsp;

&nbsp;

### Okay, seems like we already have a way of overriding then?

Yeah, kind of. There a few problems with the current/adhoc approach:

1. Difficult to track

    To be able to absorb or request changes, we should be able to track overrides.

    ```jsx
    const InputSearch = styled(TextInput)`
      background-color: #f6f6f6;
    `;
    ```

    It's not easy to distinguish occurrences of `styled(CosmosComponent)` from local uses of `styled` (without implementing some babel-parser magic)

[...]

PR #1208, opened 2018-12-04
styled-components-with-helpers

siddharthkp:
For the API suggested in margin https://github.com/auth0/cosmos/pull/1059 and overrides https://github.com/auth0/cosmos/issues/1105, we need to add a helper/utility to all presentational components

Because we use `styled-components` to create our presentational components, creating a wrapper around `styled` is a nice-ish way to maintain it.

Pros: 
1. We won't need to remember adding it in every component
2. Slightly easier to upgrade styled-components


siddharthkp:
## Added after merge:

⚠️ There is a potential breaking change here: `.attrs` doesn't work in _all_ of the cases it used to.

&nbsp;

<details>
  <summary>Click to read details</summary>
If an application is extending a cosmos component, that still works.

```js
/* works */
import { Heading } from '@auth0/cosmos'
const Title = styled(Heading).attrs({
  size: 4
})
```

But if you are using `@auth0/cosmos/styled` and creating a new component with attrs, that doesn't work:

```js
/* doesn't work */
import styled from '@auth0/cosmos/styled'
const Title = styled.input.attrs({
  size: 4
})
```


[All test cases here](https://github.com/auth0/cosmos/blob/master/core/components/_helpers/component.story.js#L40-L48)
</details>

&nbsp;

We don't promote this API anywhere, but it is possible with styled-components

ℹ️ _I didn't find any breaking cases in manage/manhattan codebase, so we should be fine_

---

Here's a snippet for the release notes:

If you are using `@auth0/cosmos/styled`, you won't be able to use `.attrs` API from `styled-components` anymore.

Example:

```jsx
import styled from '@auth0/cosmos/styled'

/* doesn't work */
const Title = styled.h1.attrs({ size: 4 })

/* do this instead */
<Heading size={4} />
```

PR #1229, opened 2018-12-06
Switch.onToggle stopPropagation of event

### LINKED ISSUE #1228: Switch needs to stopPropagation for onToggle
The `<Switch>` component's `onToggle` handling needs to stop event propagation (or pass the event to the client). 

Here's a screenshot of a problem caused by this, where the switch is on top of a card. Both are clickable. When clicking the Switch, both the switch and card's onClick are triggered:

image

astanciu:
Fixes #1228


francocorreasosa:
Thanks! 😄

PR #1234, opened 2018-12-07
Removes space when the switch label is empty

### LINKED ISSUE #1126: Make accessibilityLabel optional in Switch
This is a request that comes from the manage team: They want to remove accessibility labels. When you don't pass `accessibleLabels`, the component enforces a default `[Enabled, Disabled]`

If you try to pass an empty array, it still leaves a empty label with some padding:

image

Allow the developer to disable accessibility label

**Additional context**
We should add this as a temporary feature and enforce accessibility labels in 1.0.0. The API should probably reflect that.

andresgalante:
This PR fixes #1126 by removing with CSS the `label` element when it's empty.

🤔Might be better to remove it with react, so we don't render the node at all if it's empty.


andresgalante:
@lancelaird I did talk to Franco about changing the order of the element. In general, that's a red flag and it should be avoided since changing the visual order doesn't mean changing the order of the DOM tree. In this case, since it's a label and is linked with the checkbox with an id, it doesn't matter.

landitus:
Why is it a red flag in this case? I didn't get if we should change the order in the DOM or with CSS. In this case, the position of the label is presentational, that's why I'm suggesting doing so in CSS.

andresgalante:
@landitus exactly in this case it is not a red flag

@francocorreasosa can you merge it and open a new issue for the `order` change please?

landitus:
Good! BTW, here's the next step/issue: https://github.com/auth0/cosmos/issues/1238

PR #1244, opened 2018-12-10
Image size

### LINKED ISSUE #1106: Image get's distorted inside a container
@jcenturion Hi 👋 Heard you had some trouble with the `Image` component, can you add some info here for us to work on

**Describe the bug**

I required to have all images with the same aspect ratio. With the current implementation of `Image` I only have the possibility to set the height or the width.

**Expected behavior**

As a workaround I set the `object-fit: scale-down;` for my images so the scale is the same to all images.

**Screenshots**

**Cosmos version:**
`0.8.1`

**Additional context**
-

### LINKED ISSUE #1150: Image is not scaled down to content
**Describe the bug**
A bigger image inside the `Image` container, wont be scaled down if it does not fit.

**Expected behavior**
To be scaled down, or at least make me choose the rendering option like `scaleAspectFit` or `center` or `top`. (Here is an example of how iOS does it https://developer.apple.com/documentation/uikit/uiview/contentmode)

**Screenshots**
image

**Cosmos version:**
v0.8.1

**Additional context**
Our storybook is [here](http://auth0-manhattan.herokuapp.com/?selectedKind=Applications%7CCreation&selectedStory=form&full=0&addons=0&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)

### LINKED ISSUE #1153: Check for different image sizes and aspect ratios for Avatar
We should check how we render different types of images and sizes. This is mostly to support any user uploaded image: think rectangular images (both orientations), or very large or very small images.

A solution would be to fit images. But we can check what's the best practice for an Avatar to be bulletproof.

andresgalante:
This PR fixes #1150 by making the image fit it's container while keeping it's aspect ratio by default.

It also provides props to change the way the image fits. @siddharthkp what do you think about this API?

also solves #1106

--

fixes #1106
fixes #1150 
closes #1153


andresgalante:
@landitus check the chromatic changes, maybe we need larger assets or `scale-down` should not be the default

landitus:
Thanks for the heads up! Yes, I see two different problems:
- For the Avatar implementation, we could set the image inside of it to `contain`. 
- For the Image itself, I don't get why the image surface area gets bigger (compared to the initial stage). Not sure if this is a good thing or not, yet. 

I'm doing some tests with this PRs and multiple images to be more sure.

andresgalante:
> I'm doing some tests with this PRs and multiple images to be more sure.

Let's set it up as `max-width: 100%` and remove the props for now. We can add more options later if needed.

core/components/atoms/image/image.js
```diff
@@ -5,14 +5,14 @@ import PropTypes from 'prop-types'
...
+  height: auto;
+  max-width: 100%;
 `
 
 Image.propTypes = {
   /** URL of the image */
   source: PropTypes.string.isRequired,
-  /** Alt info for the image */
+  /** Alt text is mandatory, please read the accessibility section  */
   alt: PropTypes.string
```
siddharthkp: The way to make something mandatory with proptypes is to mark it as required. This will show up a warning on dev tools when it's missing

```js
alt: PropTypes.string.isRequired
```

Ref: https://reactjs.org/docs/typechecking-with-proptypes.html

core/components/atoms/image/image.story.js
```diff
@@ -5,18 +5,21 @@ import { Example, Stack } from '@auth0/cosmos/_helpers/story-helpers'
 import { Image } from '@auth0/cosmos'
 
 storiesOf('Image').add('url', () => (
-  <Example title="Icons">
+  <Example title="url">
```
siddharthkp: 👍

core/components/atoms/image/image.js
andresgalante: ❤️ thanks Sid

andresgalante:
> Wait, I'm a late to the party? the implementation disappeared 😅

After the new rules around generating API and props we decided not to go letting the user do whatever they want with the image size but just set a max width.

This is really strange that is not showing a diff. I'll take a look :)

core/components/atoms/image/image.md
```diff
@@ -8,3 +8,32 @@
...
+
+## Usage
+
+The imagien component can fit it's container in 3 different ways:
+
+- **contain:** The image is scaled to maintain its aspect ratio while fitting within the container content box. The image will be "letterboxed" if its aspect ratio does not match the aspect ratio of its container.
+- **none:** The image will keep it's orinigal size.
+- **scale-down:** The image is sized as if none or contain were specified, whichever would result in a smaller concrete object size. This set up keep the aspect ratio but avoids the loss of resolution.
+
+## Accessibility
```
siddharthkp: Is it to petty to point out typos 😅

andresgalante: Please do correct my typos, I suck at writing

andresgalante:
In order to merge this with a required alt text, there is a need to fix every instance of the image component in use in other components.

@siddharthkp what do you think if we merge this as is to fix the visual bug and track the alt text requirement on a separate issue?

siddharthkp:
> @siddharthkp what do you think if we merge this as is to fix the visual bug and track the alt text requirement on a separate issue?

If you want, you can move the `isRequired` change to a different PR. We shouldn't merge it without fixing the warnings (because these are internal cosmos warnings that others won't be able to fix in their code)

core/components/atoms/image/image.md
```diff
@@ -8,3 +8,28 @@
...
 ```
+
+## Usage
+
+The imagien component is scaled to maintain its aspect ratio while fitting within the container content box. The image will be "letterboxed" if its aspect ratio does not match the aspect ratio of its container.
```
siddharthkp:
```suggestion
The Image component is scaled to maintain its aspect ratio while fitting within the container content box. The image will be "letterboxed" if its aspect ratio does not match the aspect ratio of its container.
```

andresgalante:
@siddharthkp I've opened #1272 to make the alt required, I think we can move forward with this one.

@landitus ?

siddharthkp:
> Image is responsive by default, meaning it will not scale up, but it will shrink to fit the width of it's container.

I'm not sure if I understand what this means. In my head it means this:

image

> I'm inclined to declare display: block on the image. But also it would not allow it to center it easily within a content/text (with text-align: center).

Interesting! You can use a layout component to position an image in a molecule. So we might actually be fine. However, it feels a little awkward to change the property of a atomic element like ``, it would lead to a lot of confusion for sure

andresgalante:
> I'm not sure if I understand what this means. In my head, it means this

@siddharthkp I completely see what you mean, this seems like another situation were if you don't know the object fit CSS property you'd be lost in what each value means.

The size is relative to the element's box itself, and not the container element the image is nested in. That's why the image needs to have a height and width in order for `object-fit` to have an effect.

Object-fit is rarely used, and I don't think we can achieve the use-case to "scale down to fit" within a rounded shape.

I wonder if we are over-engineering the image component.

landitus:
Certainly, the fit property has very limited use, and it's optional and an enhancement for certain use cases: like the logos. You'll probably use the image as is most of the time and only go deeper when you encounter the limitations of the defaults. 

The docs have examples, but let me know if they are good enough. Probably a before and after would be more explanatory, or a demo like this: https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit

Still, want to avoid OE!

andresgalante:
👍 I am ok with the last changes, and the docs are great.

I'd add some kind of validation to the width and height strings to allow a number and a unit, and a boolean prop to disable responsive images.

andresgalante:
I've just sent a commit with the disabled responsive prop. Remember to add alt text to all the images.

landitus:
I don't know why the images are not displaying in the Resource List example / story :(

PR #1249, opened 2018-12-11
Internal: Massive find and replace for styled

siddharthkp:
Based on https://github.com/auth0/cosmos/pull/1208


andresgalante:
Is `size` a change in style or markup? or both?

PR #1250, opened 2018-12-11
[Internal] Replace margin overrides with margin prop

siddharthkp:
Built on top of https://github.com/auth0/cosmos/pull/1249

---

If this PR is too big to succeed, I can split it into smaller chunks and roll it out 
over time.

There should be 0 chromatic changes.

---

- [x] atoms
- [x] molecules 
- [x] _helpers

Left the `Stack` component alone because it needs a refactor

---

Note: The goal is not to replace all usage of margin inside cosmos.

You should use css margin while creating new components (typically atoms), but use the margin prop while composing them in molecules.

Using the margin prop inside cosmos is just a way to test out the API internally before giving it to other folks

---

## I have updated the findings in the RFC: https://github.com/auth0/cosmos/pull/1059#issuecomment-446484124

### DIFF
--- a/core/components/atoms/_overlay/overlay.js
+++ b/core/components/atoms/_overlay/overlay.js
@@ -73,7 +73,16 @@ class Overlay extends React.Component {
     if (open) {
       content = (
         <Overlay.Backdrop onMouseDown={this.handleMouseDown}>
-          <Overlay.Element contentSize={contentSize} innerRef={el => (this.contentElement = el)}>
+          <Overlay.Element
+            margin={{
+              top: 'xlarge',
+              bottom: 'xlarge',
+              left: 'small',
+              right: 'small'
+            }}
+            contentSize={contentSize}
+            innerRef={el => (this.contentElement = el)}
+          >
             {children}
           </Overlay.Element>
         </Overlay.Backdrop>
@@ -98,7 +107,6 @@ Overlay.Backdrop = styled.div`
 
 Overlay.Element = styled.div`
   width: 100%;
-  margin: ${spacing.xlarge} ${spacing.small};
   pointer-events: none;
 
   /* Since the focus trap is adding divs around the dialog box, the max width prop should be here */
--- a/core/components/atoms/_overlay/overlay.js
+++ b/core/components/atoms/_overlay/overlay.js
@@ -108,6 +108,7 @@ Overlay.Backdrop = styled.div`
 Overlay.Element = styled.div`
   width: 100%;
   pointer-events: none;
+  display: inline-block;
 
   /* Since the focus trap is adding divs around the dialog box, the max width prop should be here */
   max-width: ${props => Overlay.getSizeForOverlay(props.contentSize)};
--- a/core/components/atoms/_overlay/overlay.js
+++ b/core/components/atoms/_overlay/overlay.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import ReactDOM from 'react-dom'
 import PropTypes from 'prop-types'
-import styled from 'styled-components'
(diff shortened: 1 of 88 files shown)


andresgalante:
Hey @siddharthkp when you get to the Stack and the other scary things let me know, we can do that together

siddharthkp:
> Hey @siddharthkp when you get to the Stack and the other scary things let me know, we can do that together

Left the Stack component alone because it needs a refactor first

core/components/atoms/_overlay/overlay.js
```diff
@@ -73,7 +73,16 @@ class Overlay extends React.Component {
     if (open) {
       content = (
         <Overlay.Backdrop onMouseDown={this.handleMouseDown}>
-          <Overlay.Element contentSize={contentSize} innerRef={el => (this.contentElement = el)}>
+          <Overlay.Element
+            margin={{
```
andresgalante: I'd avoid something like this.

If a component or children of a component need margins, then it should have them on the CSS. Let it be itself.

The margin helper should be an overwrite, as a utility class, only to be used by users in the case they need to overwrite margins.

The same goes for all the other changes in this PR, Unless I am missing something I'd hate to go down this road.

siddharthkp: Unless you meant, let's still define them inside the component but with the prop instead of css - that sounds perfect

andresgalante: Let's keep them. 

Layouts remove margins of it's children. And if they are not using layouts, then let's not break anything before v1

siddharthkp: Oh I didn't understand your question correctly at first. Now I get it.

> I understand that the reason is to test out the API internally before giving it to other folks

Yep, that's the goal

> It'll make the developer experience for us worst, it'll be a problem to review PRs decreasing internal consistency of the code, as well as became a maintainability issue because we'd have to look into 2 different places to make style changes.

That's fair. We don't have to merge this PR. We can close it after looking at the implementation

An interesting question to ask here is: Does this prop make the developer experience worse for application developers as well?

andresgalante: > Does this prop make the developer experience worse for application developers as well?

I don't think so, utility classes proved to be vital. One of the design system's key aspects is that it should never assume, that means not assuming we know what the use cases specific situation they'll have to deal with.

Offering margin as a utility class is usually needed.

Other heavily used utility classes are for example font sizes, but we have that baked in text (or we should).

Plus there is always the situation when we have a new element on the page, we don't want to use a layout for it and we need a space, for example, an alert that appears. Let's take a look at this tomorrow when we test layouts further

siddharthkp: That's a good point!

I didn't realise we'll need `!important`

Breaking [example here](https://auth0-cosmos-replace-margins.now.sh/docs/#/playground?code=%250A/*%20you%20have%20access%20to%20all%20cosmos%20components%20here%20*/%250A%250Aconst%20Wrapper%20=%20styled.div%60%250A%20%20border-top:%202px%20solid%20pink;%250A%60%250A%250Aconst%20CustomHeading%20=%20styled(Heading)%60%250A%20%20margin-top:%20100px;%250A%60%250A%250Aconst%20Example%20=%20()%20=%3E%20%7B%250A%20%20return%20(%3CWrapper%3E%250A%20%20%20%20%3CCustomHeading%20size=%7B1%7D%20margin=%7B%7B%20top:%200%20%7D%7D%3EPreview%3C/CustomHeading%3E%250A%20%20%3C/Wrapper%3E)%250A%7D%250A%250Arender(Example)%250A)

Issue #1075, opened 2018-11-07
Improve hover state for icon in Button


landitus:
We have a special Button with icon that it's not changing color on hover. The button bg is changing, but the icon is not.

I'm attaching an image with an example. Note the blue changes from `link` to the `link-hover` token.

<img width="341" alt="screen shot 2018-11-07 at 15 38 45" src="https://user-images.githubusercontent.com/239215/48159221-65381e00-e2b3-11e8-8c37-231e460a9391.png">

PR #1251, opened 2018-12-11
Fix icon hover states

siddharthkp:
Fixes: https://github.com/auth0/cosmos/issues/1075

:drake-sad:

image

:drake-happy:

image


landitus:
@siddharthkp 	This one is getting uglier as we need to undo overrides for the Pagination, the pagination toolbar and the Dialog (x close button). I'm having a hard time overriding the special button, so I think we need to review this from another angle.

landitus:
Not as it is. Will close it for now as it makes more sense to go another route. 
a. Review the buttons again and try to move the overrides to the Table/Resource List
b. Add a specific appearance for it.

PR #1252, opened 2018-12-11
Fix Security issue and fix error on example when opening Applications

jvelezpo:
It fixes a npm install error with `react-dom` version and It fixes this error:

image


francocorreasosa:
Hey! Thank you for the PR! I will take a look at this and keep you posted :)

Issue #1232, opened 2018-12-07
Tooltip maxWidth


astanciu:
I think the `<Tooltip>` component could benefit from a `maxWidth` property to line wrap long texts.

![image](https://user-images.githubusercontent.com/1675025/49619656-7ac06680-f98b-11e8-805e-a76ff0df554a.png)

landitus:
Indeed, thanks @astanciu . We'll add it to next sprint.

landitus:
I'm attaching the design spec for reference. I think we should have all tooltips have a `max-width` of 260px. I've taken this number from the actual usage in Website and Manage.

<img width="790" alt="screen shot 2018-12-11 at 13 34 08" src="https://user-images.githubusercontent.com/239215/49815167-ed459380-fd49-11e8-8a5b-43a39ec0d380.png">

PR #1258, opened 2018-12-11
#1232 Add Tooltip maxWidth && fix some warnings about components with…

jvelezpo:
Fixes: 
* https://github.com/auth0/cosmos/issues/1232
* some warnings about components without keys inside a map

### DIFF
--- a/core/components/atoms/switch/switch.js
+++ b/core/components/atoms/switch/switch.js
@@ -14,9 +14,9 @@ class Switch extends React.Component {
     this.state = { on: props.on }
   }
   onToggle(e) {
-    e.preventDefault();
-    e.stopPropagation();
-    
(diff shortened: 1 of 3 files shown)


jvelezpo:
Note: 
i tried to update the Docs but i could not do it :/
ran yarn dev, and was trying to modify `tooltip.md`, and nothing ......

maybe  if i have some more time later i will get back on it

landitus:
Hi @jvelezpo 	!! Thanks for contributing! I've just reviewed the design and made a spec where the max-width is not configurable, but set to 260px. Just letting you know in case you want to fix this, but we can gladly take it from here so we can have it merged.

jvelezpo:
Im on it 👍

PR #1264, opened 2018-12-12
Makes Tabs accessible

### LINKED ISSUE #923: Tabs are not accessible
**Describe the bug**
The Tabs cannot be navigated using the keyboard by hitting tab. (the irony!)

**Expected behavior**
You should be able to move through Tab Links by using tab key, and selecting one by using the enter key

**Screenshots**
tab-tab

**Cosmos version:**
0.5.2

**Additional context**
Could be a case of non-semantic HTML output

### LINKED ISSUE #1169: Add support for Tabs in Dialog
When using a Tab component inside a Dialog, there are styling problems due to: 1) we need to achieve a edge to edge positioning, 2) Tabs are centered inside the dialog.

Provide proper support and styles to add Tabs inside the dialog. Add examples for testing and documentation for this use case. 

**Additional context**
I remember this feature was part of the Dialog PR, but we decided to add it later so we removed it.

andresgalante:
This PR closes #923 and fixes #1169 by making the tabs responsive and styling them correctly on the dialog.

Now Tabs:

- Have the correct aria tags, roles and IDs
- Handle focus correctly
- Handle keyboard interaction as described on the specs
- Have focus indications
- Are centered within the dialog
- Improves documentation


landitus:
Another comment: The focus ring on the tab content works ok when navigating with the keyboard, but I also shows when the user clicks on the area. Can we show the focus ring only when the user navigates with the keyboard? 

kapture 2018-12-14 at 12 14 49

core/components/molecules/tabs/tabs.js
```diff
@@ -129,20 +131,69 @@ class Tabs extends React.Component {
...
+Tabs.TabLink = styled.button`
+  /* Resets button browser styles */
+  background-color: transparent;
+  border: none;
+  padding: ${spacing.small} 0;
+  color: ${colors.link.default};
+  cursor: pointer;
+  border-bottom: 1px solid transparent;
+  margin-bottom: -1px;
+  line-height: 1.3;
```
siddharthkp: who dis 🤔

andresgalante: I wouldn't use the button component, they don't look the same, and if the button changes, then this component gets destroyed.
 
This is a really good case of repetition been better than the wrong abstraction.

What do you mean by "non-standard"? Cosmos has 2 line heights, 1.6 or 1.3. It'll be clearer when we finish the tokens.

siddharthkp: > even though they are the same HTML element, they are not the same component.

💯 

> Cosmos has 2 line heights, 1.6 or 1.3

wut, really? I didn't know that

andresgalante: well, now line-height is all over the place, we have components with line height `1`, or hacked to be the same as the height of the component. That is very wrong.

But hidden under the dust I could uncover that rule, and confirmed it after talking with @landitus 

I am sure it'll be clearer when we have the variable system in place since there will only be 2 global line heights tokens.

PR #1274, opened 2018-12-13
Explictly declare element ordering in switch

tehsis:
I have reworked over #1273. Instead of using an array to order elements I explicitly order the elements surrounding them by fragments assigning them to the elements variable.


francocorreasosa:
Let's merge it into #1273 and continue there :) Thanks

PR #1282, opened 2018-12-17
Layout: Row, Gallery, Columns and Page

andresgalante:
This PR is related to #1191 and introduces a bunch of important:

## Margin Resets 

It abstract the layout margin reset so it removes margins from the children of the layouts. 

This is a temporary solution until we remove margins on v1 and @siddharthkp made a comment that the name of the prop sucks. He is right, I need help finding a new name.

## Gutters

It cleans up all layouts from CSS custom properties.

## Gallery options

It introduces gallery options to create different gallery sizes. I need help with the naming of this prop too :)

## Are we friends with the margin helper API?

Yes! We've tested the layouts on our Manage implementation using the margin prop and although we had to fight specificity  we conquer it by adding a class to the children of the layout (thanks @siddharthkp for the idea).

This Pr also adds an `!important` flag to the margin helper class.

## What reminds to be done:

- @landitus decide on the number of gutters and sizes.
- Add a `...prop` to all components
- Follow the same implementation for stack and app layouts


andresgalante:
On the last commit, we've improved the name of the reset prop and made it a boolean.

examples/manage/pages/application-detail/index.js
```diff
@@ -19,39 +19,38 @@ class ApplicationDetail extends React.Component {
...
+          <ApplicationPageHeader
+            title="API Explorer Application"
+            type={{
+              name: 'Non Interactive',
+              applicationId: this.props.match.params.applicationId
+            }}
+            logo={
+              <Avatar type="resource" size="xlarge" image={ApplicationTypeImages.non_interactive} />
+            }
+            breadcrumb={{
```
siddharthkp: Does this actually render breadcrumbs :thinking:

andresgalante: I don't know, it's really strange, but it was there before unrelated to layouts

siddharthkp: It shows up as a chromatic diff because the above `<Breadcrumb/>` was removed

andresgalante: @siddharthkp I bright back the breadcrumb, can you test it out, please?

PR #1286, opened 2018-12-17
Removes line height declaration on tables

### LINKED ISSUE #1231: Fix Table row line-height
**Describe the bug**
The current table implementation has a line height of 2, which is different than the default line-height

**Expected behavior**
The Table should use the default line-height.

andresgalante:
This PR fixes #1231 by removing the line-height declaration from `th` and `td`s

It also cleans the CSS by avoiding the use of shorthands on `background`.


francocorreasosa:
@landitus can you check & merge?

PR #1308, opened 2018-12-19
Testing: Custom prop - Atoms

siddharthkp:
Testing if the component supports custom prop

To test this, I also need to grab the element by automation id, so that is getting tested as well 😄 

Skipped: Select.*


internal/docs/pages/automation-glossary.js
```diff
@@ -30,6 +30,10 @@ class AutomationGlossary extends React.Component {
                   '$(\'div[data-cosmos-key="alert"] a[data-cosmos-key="alert.dismiss"\')'
                 ]
               },
+              {
+                component: 'Avatar',
+                selectors: ['$(\'div[data-cosmos-key="avatar"]\')']
+              },
```
francocorreasosa: We should consider generating this automatically from component's code. It's going to be hard to maintain by hand

PR #1311, opened 2018-12-19
Testing: Test event handlers

francocorreasosa:
Add test event handlers to each component.


siddharthkp:
I'm glad you added the dialog test, that helps me communicate that we should not be using enzyme to test our components:

An implementation with enzyme would usually allow on the implementation details of the component. The following test (copied from the changes) will break the moment you change the structure or behavior of the dialog component.

```jsx
it('calls actions handler', () => {
    const onClick = jest.fn()
    const actions = [<Button onClick={onClick}>Something</Button>]

    const action = dialogFactory({ actions })
      .childAt(0) // FocusLock
      .childAt(0) // Dialog Element
      .childAt(3) // Footer
      .childAt(0) // Footer Actions
      .childAt(0) // First Action

    action.simulate('click')

    expect(onClick).toHaveBeenCalled()
  })
})
```

A good test should not have to be changed with change in the implementation, in fact, that's what gives confidence to change the code (only exception: unless there is a breaking change)

This is how the same test would look with react-testing-library:

```jsx
it('calls actions handler', () => {
    const body = render(<Fixture />)

    const footer = body.getByTestId('dialog.footer')
    const firstAction = footer.getAllByTestId('button')[0]

    firstAction.simulate('click')

    expect(Fixture.onClick).toHaveBeenCalled()
  })
})
```

Here's another working example from `Select`: https://github.com/auth0/cosmos/pull/1384/files#diff-969d4bc478e604be4eba4694a1a0e9c9

francocorreasosa:
@siddharthkp just brought your change to this branch so we can make the unification of this kind of test here :)

PR #1312, opened 2018-12-20
Testing: Custom prop - Molecules

siddharthkp:
Built on top of https://github.com/auth0/cosmos/pull/1308

Adding `{...props}` did change some of the snapshots, we need to review them to see if there are any bugs hidden in there


internal/test/unit/__snapshots__/tabs.test.js.snap
```diff
@@ -77,6 +80,9 @@ exports[`Tabs only render one tab title as selected 1`] = `
 exports[`Tabs only render one tab title as selected 2`] = `
 <styled.div
   data-cosmos-key="tabs"
+  id="-m0ck3d"
```
francocorreasosa: This is definitely not expected 🧐

siddharthkp: Seems about right. `id` wasn't passed along earlier, now it is: https://github.com/auth0/cosmos/pull/1312/files#diff-724c07e413e33f19689e775b3ae0da5bR90

PR #1314, opened 2018-12-20
Adds a prop to set the width to the resource list body

andresgalante:
This PR adds a prop to set the width of the resource list body as requested by @hzalaz .

Since it's an open string 💩, I'd love for @siddharthkp to pitch in before merging it

### DIFF
--- a/core/components/molecules/resource-list/item/item.js
+++ b/core/components/molecules/resource-list/item/item.js
@@ -91,9 +91,9 @@ class ListItem extends React.Component {
   render() {
     const props = this.props
     const callHandler = handler => evt => handler(evt, props.item)
-
     return (
       <ListItem.Element
+        {...props}
         draggingMode={props.draggingMode}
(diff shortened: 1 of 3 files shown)


core/components/molecules/resource-list/item/item.js
```diff
@@ -189,6 +190,8 @@ ListItem.propTypes = {
   image: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
   /** An icon to display as a thumbnail image for the list item. */
   icon: PropTypes.oneOf(__ICONNAMES__),
+  /** Defines the flex-basis of the middle column. Accepts any number and unit, for example: `40%` or `200px` */
+  bodyWidth: PropTypes.string,
```
francocorreasosa: I think we could change this for a function that expects a `positive integer number` + `%` or `px`. So we tighten a little bit more the openness of the string :)

PR #1363, opened 2018-12-27
Introduces Stack Layout

andresgalante:
@siddharthkp @landitus I really need your help with this one:

As you know `StackLayout` is a subset of flex and replaces `Stack`. Can you play with it, maybe create some stories and see how it behaves, and write your thoughts.

This I am looking to know are:
- Are we missing any option on the props? for example, distribution `space-between`
- Can you think of things you cannot build with this?
- Is the `space` prop clear enough after you use it? can you think of a better word?

@siddharthkp I'll also need your help to validate the `space` string. It should only accept numbers from 1 to 10, or the words `none` or `auto`.

Docs are missing, I am working on it today.

Thanks 🙏

### DIFF
--- a/core/components/molecules/_stack-layout/stack-layout.md
+++ b/core/components/molecules/_stack-layout/stack-layout.md
@@ -1,0 +1,37 @@
+```meta
+category: layout
+description: The StackLayout horizontally organizes its children
+```
+
+---
+
+The `StackLayout` is made of flexible items that wrap each of the stack’s children and positions them horizontally in a row. It lets you control the way they spread, space, and align.
+
+```jsx
+  <StackLayout {props}>
+    <div style={{ border: "3px dotted #666", padding: "1rem" }}>Component</div>
+    <div style={{ border: "3px dotted #666", padding: "1rem" }}>Component</div>
+    <div style={{ border: "3px dotted #666", padding: "1rem" }}>Component</div>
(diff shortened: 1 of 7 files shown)


core/components/molecules/_stack-layout/stack-layout.md
```diff
@@ -0,0 +1,37 @@
...
+
+You can nest StackLayouts or combine it with other layouts to create complex arrangements.
+
+### Do not use it for:
+
+- **Page chrome:** Use the App Frame instead for the page chrome.
+- **Create columns:** To create a column layout, use the Column Layout instead.
+- **Card List**: Use the Gallery layout instead.
+
+## Accecibility
```
francocorreasosa:
```suggestion
## Accessibility
```

siddharthkp:
Read the docs, this boy needs a lot more example to understand what the props do

andresgalante:
@siddharthkp yeap, it does. Fer suggested creating a new issue for creating examples for the stacklayout since there are so many.

andresgalante:
@siddharthkp @francocorreasosa I've opened #1416 and #1417 to track the rest of the work related to the stack layout, can we move this forward?

francocorreasosa:
Let's merge this (after at least one more approval), then let's write the docs 🤓

PR #1364, opened 2018-12-27
WIP: List component

### LINKED ISSUE #1255: [RFC] Refactor resource list
## Problem
As we build the resource list as a component we realized that it's becoming too use-case specific. On one hand, it's good because it promotes consistency, but the fact is that it's not flexible enough to cope with the needs of our author.

We ended up with a heavily modified implementation that is hard to maintain.

## Proposal

The resource list can be made out of building blocks.

We should provide enough build blocks and show examples of the different use cases specific implementation on the list component docs for an author to follow.

### What do we need?

To make sure we can reproduce the resource list we'll need to build or refine a few components:

- Create an empty list as a shell to nest any kind of list
- Finish and merge the refracting of stack layout #1113
- Improve avatar block
- Improve the button group and maybe call it a toolbar

### Margins are hitting us again!

Stack layout should take care of its children margins as any layout. But if a user wants to use a component within the list without a layout, then the margins will make spaces look wrong.

### Resource list deprecation

Since the resource list is heavily used, we might consider a longer time for deprecating the old one.

### LINKED ISSUE #1350: Improve List component
Improve list component to support handles and states from the resource list.

It should also have optional sections to resposibly display header, body, and footer. This will allow building resource list examples.

Once this component is merged we can safely remove Resource list.

andresgalante:
This is the start of an expansion to the List component.

@siddharthkp, since I suck at react, I couldn't resolve some things yet. I'll write a description of what I want to achieve and maybe we can talk about it tomorrow during mob:

- The list has an optional handle on the left that will trigger the sortable functionality
- It has an expandable option that will add an arrow on the right and expand the `Drawer` area
- It also has a `Content` area that can accept anything:
  - If the user just add a series a components, for example Stacks, each one of them will land in a new row within the Content.
  - If they chose to use our internal structure of header, body (option) and footer (optional), they'd get a reponsive structure much like resource list.

Questions are:

- What's the best strategy to create 2 "child" areas one for the Content and one for the Drawer
- How can I fix my code to make the Content child to show what it should show and not a strange repetition of items :)
- And finally, for accecibility we should have an id on the content. How can we make the author do that correctly?

When merge it'll fix #1255 and close #1350


andresgalante:
Missing from this PR are examples on the documentation, but the code is ready for reviews

core/components/molecules/_stack-layout/stack-layout.js
```diff
@@ -82,7 +82,7 @@ StackLayout.propTypes = {
   /** Horizontally distribute the components within the stack. applyes to `justify-content` */
   distribution: PropTypes.oneOf(['spaceBetween', 'center', 'start', 'end']),
   /** Flex wrap */
-  wrap: PropTypes.bool,
+  wrap: PropTypes.oneOf(['true', 'false']),
```
cristiandouce: @francocorreasosa Question: Why this change?

From a api usage point of view it'd feel more intuitive to use a `true,false` as booleans than `"true", "false"` as strings...

example

```jsx
<Stack wrap={true} >..</Stack>
```

vs

```jsx
<Stack wrap="false">..</Stack>
```

francocorreasosa: Yeah, you're right. This is currently mapping 1:1 to HTML but we can make the change inside the component for sure.

core/components/molecules/list/list.story.js
```diff
@@ -1,51 +1,287 @@
...
+            <Button icon="copy" />
+            <Button icon="delete" />
+          </ButtonGroup>
+        </StackLayout>
+      </List.Item>
+
+      <List.Item>
+        <StackLayout distribution="spaceBetween">
+          <Heading size={4}>Heading</Heading>
+          <ButtonGroup compressed>
```
landitus: No need for `compressed` here as the usual pattern for actions in lists uses buttons with spacing between them.

landitus:
We are missing two things from what I see:
- and on click behavior for the whole row: this is meant to make the whole row clickable to access the detail page
- styles for the open drawer state. I can get them to you

Anyway, this can be done in another PR if you wish.

andresgalante:
> styles for the open drawer state.

When it's open the whole thing should be grey, right? same styles as hover.

andresgalante:
@francocorreasosa can you help with this please?

- and on click behavior for the whole row: this is meant to make the whole row clickable to access the detail page
- styles for the open drawer state. I can get them to you

For the second one I'd need to add a class to the `<li>` when the drawer is open so we can paint the whole thing grey.

PR #1371, opened 2019-01-02
Use react-popper for tooltips

siddharthkp:
Fixes https://github.com/auth0/cosmos/issues/1367 and https://github.com/auth0/cosmos/issues/951

PR #1372, opened 2019-01-02
Accessible tooltip

siddharthkp:
Based on top of https://github.com/auth0/cosmos/pull/1371 (Review after merging that)

---

Followed this spec: https://www.w3.org/TR/wai-aria-practices-1.1/#tooltip

- [x] The element that serves as the tooltip container has role tooltip.
- [x] The element that triggers the tooltip references the tooltip element with `aria-describedby` (`props.id` is expected)
- [x] Focus stays on the triggering element while the tooltip is displayed.
- [x] Tooltip never gets focus
- [x] If the tooltip is invoked with mouseIn, then it is dismissed with on mouseOut.
- [x] If the tooltip is invoked when the trigger element receives focus, then it is dismissed when it no longer has focus (onBlur).
- [x] Hitting escape dismisses the Tooltip (including in cases with `props.defaultVisible`)

---

This probably breaks the use case with an icon: https://inclusive-components.design/tooltips-toggletips/#inclusivetoggletips

Need @andresgalante help to understand it better


andresgalante:
Great job! 

There are 2 scenarios for tooltips and I am not sure how we can handle them:

The idea is to associate the tooltip text with the element that triggers the tooltip: 

- We should use `aria-labelledby` if the tooltip provides a label for the element. For example, an icon of an envelope that has a tooltip that says "email".

- On the other hand, we should use `aria-describedby` if the element already has an accessible label that is different from the tooltip text. For example, a button that has a label, and the tooltip provides more information.

## Toggle tips and other kind of pop overs

You are right "Toggle tips" or "Popovers" are a different use case. 

Sometimes popovers include an "x" icon just like modals, to close them. that's why they receive a role of `dialog` instead.

For the icon toggle tips, since they are the same design and they just change the opening/closing behavior I think we can create a boolean prop that changes the behavior. If it was a different design, then I'd made them a different component. 

What do you think?

siddharthkp:
Oops, we talked about this but I forgot to document it here for others to read

>

andresgalante:
100% agreed, let's make the toggle tip a new component.

landitus:
There are a few things that we need to fix:

- Tooltips should be separated from the element. Let's use the smallest spacer (4px)
- Tooltips are slightly off-centered from the element.   

---
I see what you mean about the Toggle tooltip, and we can consider building it. It appears often, indeed. 
---
BTW, seems story https://www.chromaticqa.com/component?appId=5aba7f6a07f95800202658ff&name=Tooltip&buildNumber=1839&specNumber=6&viewport=1200&browser=chrome has a chromatic bug, as it looks weird there.

Issue #1267, opened 2018-12-12
Label id uses random generated value and breaks snapshot tests


francocorreasosa:
```diff
    @@ -264,11 +264,11 @@
            <div
              className="sc-iyvyFf bhhnDL"
            >
              <label
                className="sc-dqBHgY hqeZQu"
-               htmlFor="add_same-origin_header_to_authentication_endpoints-ubv6jvm3"
+               htmlFor="add_same-origin_header_to_authentication_endpoints-vymhlvxa"
              >
                Add Same-Origin header to authentication endpoints
              </label>
            </div>
            <div
```

tehsis:
I have also reported this on: https://github.com/auth0/cosmos/issues/1268

siddharthkp:
Fixed in #1374

PR #1374, opened 2019-01-03
Alternate implementation for uniqueId

siddharthkp:
Will wait for @francocorreasosa's review

----

Story time!

In certain components, we want to create a relation between
2 html elements

examples:

1. form field labels

    ```jsx
    <Form.Field>
      <Label for="my-input">field</Label>
      <TextInput id="my-input"/>
    </Form.Field>
    ```

2. tabs

    ```jsx
    <Tab.Link id="first-tab">Tab 1</Tab.Link>
    <Tab.Item id="first-tab-item" aria-labelledby="first-tab">...<Tab.Item>
    ```

We try to implement this without asking to specify an id/label for each of these items

A naive implement (the previous one) would be to generate random unique ids
and use them. However this approach breaks snapshot tests because the id is
not consistent across multiple test runs.

This can be solved in user land by mocking this function
jest.mock('path_to_this_file', () => () => '-m0ck3d')

Another solution could be to generate IDs based on an "pure"-ish algorithm,
which gives the same id as long as other environment factors are not changed
1. In an isolated environment, the IDs are consistent across each run
2. In a composed applications, the IDs change but are still unique

The implementation, it could just be an increasing integer.

Related to #1267


francocorreasosa:
I'm scared about having a global counter, any ID may change due to a tiny change in one corner of an app. Unfortunately I don't have a better solution in mind, we should try to use this function as less as we can. 

We should aim to give Cosmos' users the chance to explicitly define an ID in their components and use this as a fallback. ([as we do in Tabs](https://github.com/auth0/cosmos/blob/master/core/components/molecules/tabs/tabs.js#L195)).

siddharthkp:
> I'm scared about having a global counter, any ID may change due to a tiny change in one corner of an app. Unfortunately I don't have a better solution in mind, we should try to use this function as less as we can.

What's the problem with that 🤔 All of this happens with the app as a whole. Not when components are rendered in isolation (like in tests)

francocorreasosa:
Oh I see what you mean, in that case we're fine. Let's merge this.

PR #1376, opened 2019-01-03
[Proposal] Support integration with client-side routing

nkohari:
Hey guys! 👋  I wanted to propose something, and felt like it was easier to illustrate by implementing it and opening a PR. 😄 

In Manhattan we've been having a problem where `<Link>` components with `href` props (that is, normal hyperlinks) cause a full page reload when a user clicks them. To fix this, we need to catch the click event and pass it to the client-side router. We can handle that with an `onClick` handler, but that has a couple of problems:

1. Since they aren't real links, UX suffers, since users can't modifier-click (eg. meta-click) to open the link in a new tab or window.
2. Without manually setting tab stops, we might run into issues with accessibility.

Worse, some Cosmos components contain embedded `Link` elements, and it's not always clear how to override the click handlers for them.

At the same time, Cosmos has to stay router-agnostic, and the normal browser behavior of a full page refresh is a sensible default. For apps like Manhattan, though, we need a way to externally tell Cosmos what we'd like to happen when a user clicks a `Link` with an href.

This PR adds a `<Provider>` component, which uses React context to expose configuration information to Cosmos components. Then, the `Link` component can consume this context, and if an `onNavigate` function is provided, it can call it. For example:

```
const onNavigate = (event, href) => {
  history.go(href); // pass the navigation event on to a client-side router
  event.preventDefault();
};

const config = { onNavigate };

const Example = () => (
 <Provider config={config}>
   <Link href="/path/to/something">click me</Link>
 </Provider>
);
```

When the `Link` in this example is clicked, instead of the browser navigating to `/path/to/something`, instead the `onNavigate` function from the `Provider` will be called.

If someone passes an `onClick` handler to a `Link`, it will override the `onNavigate` function. Likewise, if the `Link` doesn't have an `href`, `onNavigate` won't be called.

We kicked around a couple of other proposals for APIs, but I felt like using context had these benefits:

1. It's the most idiomatically-correct way to inject configuration in React
2. It provides an extension point which might be useful for other things in the future (theming?)

WDTY? 🤔 I'm happy to add docs to this PR if you guys are happy with the API.


siddharthkp:
@nkohari Naaaaate hi 😊

I skimmed through this right now, it needs a better look. I'll send some time on it tomorrow (it's late here) 👍

nkohari:
Yeah, that's true. That would have to be solved in `onNavigate`. Right now we have a PR for a click handler that tries to do this, but adapting it to work with this PR, it would be something like this:

```
const onNavigate = (evt, href) => {
  const comboKey = evt.metaKey || evt.altKey || evt.ctrlKey || evt.shiftKey;
  if (!comboKey) {
    evt.preventDefault();
    history.go(href); // or whatever we need to do to move the client-side router
  }
};
```

siddharthkp:
> const comboKey = evt.metaKey || evt.altKey || evt.ctrlKey || evt.shiftKey;

That's the one I'm confused about. If we have to do this for all `Link`s, we might as well do this in the system (with tests, etc)

nkohari:
Sure, I think that'd be reasonable too; it's just a question of what control you want to mandate within Cosmos and what control you want to delegate to the app developer.

nkohari:
Also keep in mind there's only a single instance of `onNavigate` per app, so it isn't that we have to repeat that code that handles the meta-click on every `Link` or anything.

siddharthkp:
> it's just a question of what control you want to mandate within Cosmos and what control you want to delegate to the app developer.

yep. I feel this is going to be a cross product concern, that we can absorb it in the system.

> Also keep in mind there's only a single instance of onNavigate per app, so it isn't that we have to repeat that code that handles the meta-click on every Link or anything.

oh yeah, that's a big one.

siddharthkp:
WIP, partial comment coz I'm scared of losing progress

Hi 👋 

Here's how I would redefine the problem (let me know if I'm wrong)

Client side routers come with their own `Link` component. In the application, we want this link to look (and behave) just like a `Link` from cosmos.

In it's current shape, the `Link` component [from cosmos](https://github.com/auth0/cosmos/blob/master/core/components/atoms/link/link.js) only has enough code to style it.

One way to borrow these would be to just apply the same styles to the `Link` from the router


```jsx
import { Link as RouterLink } from 'react-router5'

const Nav = () => (
  <nav>
    <RouterLink routeName="page1">Page 1</RouterLink>
  </nav>
)
```



The expectation from the `Link` component from cosmos is: 

1. Consistent look
1. Respect accessibility when used inside a molecule
1. Have the right automation attribute



In Manhattan we've been having a problem where <Link> components with href props (that is, normal hyperlinks) cause a full page reload when a user clicks them.

siddharthkp:
> What concerns do you have about the original proposal? Is it the use of context, or something else?

I should have started with that, sorry.

I really like the use of context to pass config. In fact, we probably should move [`COSMOS_DISABLE_RESETS`](https://auth0-cosmos.now.sh/docs/#/usage#usage-with-styleguide) to the provider as well.

The part I don't like is the `onClick` handler because we need to re-implement all the features of an anchor tag (1. modifiers, 2. target, 3. usage with keyboard, 4. more?)

I would like to use an anchor tag in it's pure form if we can.

> For the first solution, we'd have to either ensure that we apply the styles on each individual Link in the application, or create our own reusable component (eg. ManhattanLink) that applies the styles.

Yep, that's common to all the solutions. 

I see an `onNavigate` handler (with context) and `ManhattanLink` as the same level of abstraction/indirection

There might be another approach here to make the `as` more global with context:

### Approach 3

```jsx
import { Provider, Link } from '@auth0/cosmos'
import { Link as RouterLink } from 'router5'

<Provider components={ Link: RouterLink }>
  <Link routeName="/page2" target="_self">click me</Link>
  /* routeName is a router prop which will be proxied through  */
</Provider>
```

nkohari:
> The part I don't like is the onClick handler because we need to re-implement all the features of an anchor tag (1. modifiers, 2. target, 3. usage with keyboard, 4. more?)

No, Cosmos would still render a regular anchor with an `href`, so `target` and tab stops would work fine. It's true that we'd have to handle modifiers in the `onNavigate` function. Client-side router `Links` do the same thing; for example, here's Router5's:

https://github.com/router5/router5/blob/master/packages/react-router5-hocs/modules/BaseLink.ts#L87

We could put that logic inside of Cosmos (meaning `onNavigate` is only called only on regular keypresses and not on modifier-clicks), or you could leave that up to the app developer to decide. Either way it's not a lot of complexity.

siddharthkp:
You don't like the component customisation with context? ⬇️ 

```diff
render(
-  <Provider config={ onNavigate }>
+  <Provider components={ Link: RouterLink }>
    <App/>
  </Provider>
)
```

siddharthkp:
@nkohari 👋

siddharthkp:
@cristiandouce Hi 👋 

Can you review this proposal and the suggestions in the comments. [Approach 3](https://github.com/auth0/cosmos/pull/1376#issuecomment-451952712) is what I'm leaning towards, let me your opinion.

PR #1393, opened 2019-01-08
Fixed rendering of label for Radio and Form.Radio (with tests)

landitus:
Fixes https://github.com/auth0/cosmos/issues/1394

Added missing stories for testing Radio and Form.Radio. Note that these tests don't look good because there's a visual regression, and are meant to be used to test the fix and to avoid future regressions.

I will approve the chromatic tests so we can merge.


siddharthkp:
@landitus Fixed in this PR

landitus:
Thanks @siddharthkp 	!! Just changed a margin so we are ready to merge!

PR #1404, opened 2019-01-09
Do not enforce totals on the Pager component

### LINKED ISSUE #1219: Pager component displays undefined
When you don't provide the total number of items, the **Pager** component displays `undefined` in the UI. It also doesn't make sense when the number provided is 0 or negative
Pager could allow to not necessarily pass all options. Example: Items could be optional when loading items from a cursor (not knowing the extent of items at all).

**Describe alternatives you've considered**
https://manage.auth0.com/#/logs

**Additional context**

> **Cosmos displays undefined when total amount of items is missing**

image

> **Total number of items doesn't make sense here**

image

image

francocorreasosa:
Resolves #1219


landitus:
Speaking with @francocorreasosa , in the case we don't have totals information, we shouldn't force to display items or anything related to it. I would just display the Page buttons at both sides.

Issue #1405, opened 2019-01-09
Add missing icons


landitus:
There are missing icons in Cosmos that are used across Manage. 

This is a preliminary inventory. Number corresponds to here: http://styleguide.auth0.com/#/resources

- [x] 110 
- [x] 115 
- [x] 142 
- [x] 143 
- [x] 148 
- [x] 154 
- [x] 168 
- [-] 171 X Use rotate
- [x] 173 
- [x] 178 
- [x] 187 
- [x] 188 
- [x] 243 
- [x] 256 
- [-] 257 X
- [x] 263 
- [-] 264 X Use delete
- [-] 266 X Use pencil
- [x] 274 
- [x] 285 
- [x] 289 
- [x] 292 
- [x] 299 
- [x] 300 
- [-] 327 X Use settings or usage
- [x] 329 
- [-] 330 X Use settings
- [-] 333 X Use settings
- [x] 338 
- [x] 339 
- [x] 342 
- [x] 347 
- [x] 354 
- [x] 355 
- [x] 359 
- [x] 362 ? Not sure
- [-] 368 X Use bars
- [x] 372 
- [x] 374 
- [-] 389 X Use delete
- [-] 434 X Use reload
- [x] 436 
- [-] 446 X Use download
- [x] 458 
- [x] 460 
- [x] 461 
- [x] 462 
- [x] 463 
- [x] 466 
- [x] 470 
- [x] 471 
- [x] 473
- [x] 478
- [x] 480
- [x] 489
- [x] 499
- [x] 519
- [x] 521
- [x] 522
- [x] 541
- [x] 546
- [x] 606
- [x] 608 ?
- [x] 619
- [x] 661
- [-] 669 Use copy
- [x] 697
- [x] 371
- [x] 375
- [x] 753
- [x] 754
- [x] 755
- [x] 761
- [x] 764
- [x] 774
- [x] 778
- [x] 799
- [-] 825 Use brand-microsoft

PR #1406, opened 2019-01-09
Added missing icons

### LINKED ISSUE #1155: Separate Icons types in the documentation
We now have UI icons and brands in the same grid. This makes it difficult to browse. Also, it makes sense to keep them separated.

Let's have 2 different grids: one for UI Icons and other for Brand Icons. We could add a title for each grid.

**Additional context**
Let's make sure the Search works after this change.

landitus:
Closes https://github.com/auth0/cosmos/issues/1405
Closes https://github.com/auth0/cosmos/issues/1415
Closes #1155 

This PR also improves the display of icons in the Docs:
- Improved design for better legibility
- Improved search: you can now search icons by their number if you are migrating from Styleguide


core/icons/aliases.json
```diff
@@ -1,10 +1,49 @@
 {
-  "delete": ["trash", "remove"],
+  "database": ["110"],
```
francocorreasosa: Adding budicon numbers to improve searchability in docs 🤩 very clever

landitus:
I see a weird issue in the Icon docs page:
- In the Props Table, the select for name is not showing all the icons available, for example: `application-plus` is not showing.

screen shot 2019-01-11 at 00 11 24

core/icons/aliases.json
siddharthkp: Don't remove the old ones though 😢

landitus: Nope! There are some re-arrangements on the file, since it's manual. It's error prone ... but I'm keeping them all (but moving some defs around the file)

PR #1411, opened 2019-01-10
Replaces the default value of the form label to be an empty string

### LINKED ISSUE #1410: Remove default value from forms labels
Form labels have a [default value of "Form Label"](https://github.com/auth0/cosmos/blob/master/core/components/molecules/form/field/field.js#L95).

If the author doesn't define a `label` prop, the value should be blank.

andresgalante:
This PR fixes #1410 by replacing the default value of the form label to be an empty string.


siddharthkp:
The change looks fine, you might want to update the stories and snapshots

PR #1413, opened 2019-01-10
Fix rounded corners for button group with tooltip

siddharthkp:
Found a big in master, fixed it on top of https://github.com/auth0/cosmos/pull/1371


siddharthkp:
Fixed in https://github.com/auth0/cosmos/pull/1371

PR #1429, opened 2019-01-11
Support buttons as actions in Page Header

### LINKED ISSUE #1265: PageHeader primaryAction does not support `disable` prop
I need to disable the primary action of a PageHeader until something is ready in my app.

It would be cool to support all button props in `PageHeader.primaryAction` and `PageHeader.secondaryActions`.

**Describe alternatives you've considered**
-

**Additional context**
-

francocorreasosa:
Closes #1265

### DIFF
--- a/core/components/_helpers/button-prop.js
+++ b/core/components/_helpers/button-prop.js
@@ -1,0 +1,16 @@
+import React from 'react'
+import Button from '../atoms/button'
+export default (prop, overrides = {}) => {
+  if (!prop) return null
+  if (prop.type === Button) return React.cloneElement(prop, overrides)
+
+  if (typeof prop === 'object') {
+    return (
+      <Button icon={prop.icon} onClick={prop.handler} {...overrides}>
+        {prop.label}
+      </Button>
+    )
+  }
+
+  throw new Error('button prop must be object or type Button')
+}
--- a/core/components/_helpers/button-prop.js
+++ b/core/components/_helpers/button-prop.js
@@ -1,13 +1,14 @@
 import React from 'react'
 import Button from '../atoms/button'
-export default (prop, overrides = {}) => {
-  if (!prop) return null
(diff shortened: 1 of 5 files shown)


francocorreasosa:
@siddharthkp 	Can you review & merge if everything's alright? thanks 🙇

core/components/_helpers/button-prop.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import Button from '../atoms/button'
+export default (prop, overrides = {}) => {
```
siddharthkp:
```diff
- prop
+ props
```

core/components/_helpers/button-prop.js
```diff
@@ -0,0 +1,16 @@
...
+  if (!prop) return null
+  if (prop.type === Button) return React.cloneElement(prop, overrides)
+
+  if (typeof prop === 'object') {
+    return (
+      <Button icon={prop.icon} onClick={prop.handler} {...overrides}>
```
siddharthkp: We should proxy through all of the props (not just icon and onClick)

```jsx
<Button icon={props.icon} onClick={props.handler} {...props} {...overrides}>
```

core/components/_helpers/button-prop.js
francocorreasosa: It is actually one prop (either `primaryAction` or `secondaryAction` in the PageHeader case)

siddharthkp: then let's rename it to `action`

`prop` doesn't communicate the intent

core/components/_helpers/button-prop.js
francocorreasosa: Should we? This way of using the props is being deprecated, maybe using other buttons props can be an incentive to move to the new API.

francocorreasosa:
Merging this as-is, and then we can add the remaining case if we consider is needed.

PR #1431, opened 2019-01-14
Base for Tokens v2

siddharthkp:
@landitus you might need this


siddharthkp:
Removed the usage example

andresgalante:
thanks Sid :)

PR #1434, opened 2019-01-15
Add replacement to deprecation warning

siddharthkp:
Fixes https://github.com/auth0/cosmos/issues/1388

There was just one place where we didn't have a replacement: `ResourceList.Item`

Found a rogue file that isn't imported anywhere: https://github.com/auth0/cosmos/blob/master/core/components/_helpers/icon-prop.js, @francocorreasosa can we delete this?

Unrelated change: Changed the first item in `Button` prop `type` to `submit` so that it shows up first (that's the html default)

### DIFF
--- a/core/components/atoms/button/button.js
+++ b/core/components/atoms/button/button.js
@@ -311,7 +311,7 @@ Button.propTypes = {
   success: PropTypes.bool,
 
   /** Type of button */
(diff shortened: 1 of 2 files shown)


andresgalante:
The new list component will be a suitable replacement for resource list once it's merge and we build all the examples

francocorreasosa:
@siddharthkp we can remove it, yes.

PR #1439, opened 2019-01-16
Automatically add id inside form fields

siddharthkp:
Adds for + id pair for all form fields. (Fixes https://github.com/auth0/cosmos/issues/1438)

I also realised that clicking the label in the old API worked only for TextInput, TextArea and Select (not Switch, Radio, Checkbox, etc.)

You can test them here:

| API        | Stories |
| ------------- |-------------|
| Production new API | [all fields](https://auth0-cosmos.now.sh/sandbox/?selectedKind=Form&selectedStory=all%20the%20fields&full=0&addons=1&stories=1&panelRight=0) |
| Production Old API | [switch](https://auth0-cosmos.now.sh/sandbox/?selectedKind=Form&selectedStory=switch%20field%20-%20old%20API&full=0&addons=1&stories=1&panelRight=0) |
| This branch | [all fields](https://auth0-cosmos-sid-form-field-a11y.now.sh/sandbox/?selectedKind=Form&selectedStory=all%20the%20fields&full=0&addons=1&stories=1&panelRight=0) |

---

Please read the long story in the code comments and review comments: 
https://github.com/auth0/cosmos/pull/1439/files

This is definitely peak complexity. We need to decide if we want to own this complexity up or pass the responsibility to add `for + id` to the user instead.

---

I'm not sure what's the expectation when you click a label with a Radio and Checkbox groups 🤔


internal/cosmos-scripts/configs/webpack.development.config.js
```diff
@@ -17,7 +17,8 @@ module.exports = {
   },
   devtool: 'cheap-module-source-map',
   devServer: {
-    contentBase: path.resolve(process.cwd(), 'public')
+    contentBase: path.resolve(process.cwd(), 'public'),
+    disableHostCheck: true
```
siddharthkp: Temporary measure to disable host checks, we should upgrade webpack-dev-server (Reference: https://github.com/webpack/webpack-dev-server/issues/1604)

core/components/molecules/form/field/field-input.js
```diff
@@ -0,0 +1,122 @@
...
+    const componentName = firstChild.type.displayName
+    if (cosmosInputWhitelist.includes(componentName)) firstChildProps.id = id
+
+    /*
+      If it's a cosmos component, it uses styled-components,
+      which means we need to use innerRef
+
+      Note: This will give an error for a non cosmos element :/
+    */
+    firstChildProps.innerRef = this.firstChildRef
```
siddharthkp: The fact that this gives an error if you don't use a styled-component is stupid. We might be able to come up with a fix for it. 

I didn't invest time in this because this is [solved](https://www.styled-components.com/docs/api#deprecated-innerref-prop) in styled-component@v4

siddharthkp:
After you read the code comments, read this:

There's another implementation with context which is far cleaner but has a semantic bug with multiple fields.

core/components/molecules/form/field/field-input.js
```diff
@@ -0,0 +1,116 @@
...
+  2. After render:
+
+     On componentDidMount, we find the first input element and set the id attribute
+     if it wasn't already set.
+
+     This means the rendered output is initially not accessible, but is made
+     accessible progressively.
+*/
+
+const cosmosInputWhitelist = ['TextInput', 'TextArea', 'Select']
```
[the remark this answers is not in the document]
francocorreasosa: We could use a generic field decorator for this to prevent this list from getting out of sync, so we declare on each input component implementation whether is in the whitelist or not.

+1 on making this progessively

core/components/molecules/form/field/field-input.js
```diff
@@ -0,0 +1,129 @@
...
+      Otherwise we perform a progressive enhancement and attach an id
+      after render = componentDidMount.
+
+      We keep track of the first child by attaching a ref to it.
+    */
+
+    let firstChildProps = {}
+
+    const componentName = firstChild.type.displayName
+    if (cosmosInputWhitelist.includes(componentName)) firstChildProps.id = id
```
francocorreasosa: We could check directly for the type and not the name, so we don't have to worry about maintaining `displayName`s (i'm not 100% about this though, we could check if it works as expected)

andresgalante:
I am going to paste some comments about the ids here, so I dont lose them:

I think deque mentions this scenario in their online course… that you have to pay for. I’ll check what they recommend

```
<label for="ssn1b">Social Security Number <span class="offscreen">first three digits</span>:
</label>
<input type="text" size="3" maxlength="3" id="ssn1b"> -
<label for="ssn2b" class="offscreen">Second two digits:</label>
<input type="text" size="2" maxlength="2" id="ssn2b"> -
<label for="ssn3b" class="offscreen">Last four digits:</label>
<input type="text" size="4" maxlength="4" id="ssn3b">
```

A couple of notes about this method:

-This method is more backward-compatible than the method of using aria-label, so if you’re trying to support older browsers and older screen readers, this is the best method.
-It also has the advantage that none of the screen readers will repeat any of the labels, so it sounds perfect in all screen readers.

```
<label for="ssn1">Social Security Number:</label>
<input type="text" size="3" maxlength="3" id="ssn1" 
aria-label="Social Security Number first three digits"> -
<input type="text" size="2" maxlength="2" aria-label="middle two digits"> -
<input type="text" size="4" maxlength="4" aria-label="last four digits">
```

Take a close look at the code above. There are a few things worth pointing out:

- The first field actually uses a regular <label> tag (with id and for). The reason for using the regular <label> tag is to make the text label clickable, so that when you click on the text the focus goes into the first form field of the group. This helps people who may have less manual dexterity, due to tremors or other conditions.
- The first field also has an aria-label value in addition to the <label> tag. Why both? Because we want to specify that this field is for only the first three digits. If we just used the label “Social Security Number,” the blind person wouldn’t know that the field is meant for only 3 digits, so the person will try to put in the full social security number, which would be wrong for this field.
- Will screen readers read both the <label> tag and the aria-label value? JAWS and NVDA will read only the aria-label value. VoiceOver will read both, which is less than ideal, but not a complete deal-breaker.

```
<fieldset>
<legend>Social Security Number</legend>
<label for="ssn10" class="offscreen">First three digits</label>
<input type="text" size="3" maxlength="3" id="ssn10"> - 
<label for="ssn20" class="offscreen">Second two digits:</label>
<input type="text" size="2" maxlength="2" id="ssn20"> - 
<label for="ssn30" class="offscreen">Last four digits:</label>
<input type="text" size="4" maxlength="4" id="ssn30">
</fieldset>
```

Notes about this method:

- This works well in all screen readers and is very backward-compatible.
- One drawback is that <fieldset> elements and <legend> elements can be tricky to style, compared to regular text, so some designers will prefer the other methods that allow for somewhat more CSS flexibility. Even so, modern browsers allow designers to turn off the <fieldset> border and to position the <legend> to the left of the fields. It just takes more effort. Older browsers are less flexible for styling.
- Another drawback is that some screen readers (JAWS) will repeat the <legend> text with every field within the <fieldset>. In this example, users will hear “Social Security Number” along with the labels for the three form fields. It can be annoying, but it’s not a deal-breaker either, unless perhaps if the <legend> is very long. Users can also hit tab to go to the next field to avoid listening to the entire <legend> again, or hit the control (CTRL) key to make the screen reader stop talking.

francocorreasosa:
@siddharthkp We have some Chromatic changed we should check

siddharthkp:
@francocorreasosa Just checked the diff, those are intentional. I removed the outdated `defaultSelected` prop in select and that fixed a few default values

core/components/molecules/form/field/field-input.js
siddharthkp: Removed whitelist

siddharthkp:
@francocorreasosa I've approved the chromatic tests (see previous comments)

Can you help check if this plays well with https://github.com/auth0/cosmos/pull/1483

internal/cosmos-scripts/configs/webpack.development.config.js
francocorreasosa: Opened an issue for that: #1440

francocorreasosa:
Sure will do.

francocorreasosa:
@siddharthkp It works, tested it in `Field#all the fields` story in the `franco-test-1439-1483` branch. Test commit: https://github.com/auth0/cosmos/commit/6872fec359bd59d3d9f34db93bb8da64100bf448.

francocorreasosa:
@siddharthkp is this one ready? can you merge it if so? thanks 🙇

Issue #971, opened 2018-10-05
Server rendering support for Cosmos


francocorreasosa:
_(vía @davidpatrick)_

**Describe the bug**
`document` is not defined on the server so the code breaks here:
https://github.com/auth0/cosmos/blob/2a55d0f8ad00fdff2deb3ac27f0a547ecbd49487/core/components/_helpers/globals.js#L10

francocorreasosa:
Re-labeling this as bug since this should work by default and we broke it.

PR #1456, opened 2019-01-23
Fix SSR (#971)

nicogarcia:
#### Description
SSR is currently broken because we're making use of `document` to include the style resets that cosmos requires.

To solve that problem my approach was to add a flag to avoid the automatic inclusion of styles in the Document and to export the minimum and recommended resets so they can be applied differently according to the platform.

Example:
```js
// with COSMOS_MANUAL_RESETS=true

import { resets } from '@auth0/cosmos'

insertDifferentlyAtHead(resets);

// or for Next.js

export default () => (
  <div>
    <style jsx global>{resets}</style>
  </div>
)
```

Let me know if you think the approach makes sense! 🙂 
(I know some naming, export/import decisions are debatable, please comment if you disagree with them)

#### Related Issue
#971 

#### Motivation and Context
To have SSR

⚠️ Not tested yet

### DIFF
--- a/core/components/_helpers/globals.js
+++ b/core/components/_helpers/globals.js
@@ -1,30 +1,7 @@
 import { fonts, misc } from '@auth0/cosmos-tokens'
+import { insertAtHeadStart } from './dom'
 
-let includeGlobals = true
-
-if (process && process.env && process.env.COSMOS_DISABLE_RESETS) {
-  includeGlobals = false
-}
-
-const insertAtTheStart = styles => {
-  let tag = document.getElementById('cosmos-globals')
-
-  if (tag) {
-    tag.innerHTML = styles
-  } else {
-    tag = document.createElement('style')
-    tag.type = 'text/css'
-    tag.id = 'cosmos-globals'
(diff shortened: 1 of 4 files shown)


core/components/_helpers/globals.js
```diff
@@ -1,30 +1,7 @@
 import { fonts, misc } from '@auth0/cosmos-tokens'
+import { insertAtHeadStart } from './dom'
```
siddharthkp: Nice abstraction 👍

siddharthkp:
Hi 👋 

Just a ⚠️ warning here, adding a check will stop the app from breaking but this isn't the full SSR support

We need some more work (can be done in an independent pull request) to render styles from cosmos in the server. Otherwise, you would get a flash of unstyled content until the client JS bundle arrives.

Ref: https://www.styled-components.com/docs/advanced#server-side-rendering

francocorreasosa:
@nicogarcia I'm going to merge this into an internal branch so our CI can run normally.

francocorreasosa:
Let's continue in #1457

PR #1457, opened 2019-01-24
Fix SSR (#971) (#1456)

francocorreasosa:
Original PR by @nicogarcia 

#### Description
SSR is currently broken because we're making use of `document` to include the style resets that cosmos requires.

To solve that problem my approach was to add a flag to avoid the automatic inclusion of styles in the Document and to export the minimum and recommended resets so they can be applied differently according to the platform.

Example:
```js
// with COSMOS_MANUAL_RESETS=true

import { resets } from '@auth0/cosmos'

insertDifferentlyAtHead(resets);

// or for Next.js

export default () => (
  <div>
    <style jsx global>{resets}</style>
  </div>
)
```

Let me know if you think the approach makes sense! 🙂 
(I know some naming, export/import decisions are debatable, please comment if you disagree with them)

#### Related Issue
#971 

#### Motivation and Context
To have SSR

⚠️ Not tested yet

---------------------
- Add env variable to flag manual style resets
- Export minimum and recommended style resets
- Export function to apply resets manually with ability to provide
custom applying funcion


francocorreasosa:
@siddharthkp Sid, we can render global styles with `injectGlobal` on the server, right?

core/components/_helpers/globals.js
```diff
@@ -113,5 +87,15 @@ if (includeGlobals) {
       -webkit-font-smoothing: antialiased;
       -moz-osx-font-smoothing: grayscale;
     }
-  `)
-}
+`
+
+const includeGlobals = process && process.env && process.env.COSMOS_DISABLE_RESETS
```
siddharthkp: Hi 👋

There's a flag for this behavior already: COSMOS_DISABLE_RESETS

Docs: auth0-cosmos.now.sh/docs/#/usage#usage-with-styleguide

Unless I'm missing something here, we can re-use the same flag

siddharthkp:
@francocorreasosa We can..

> Just a ⚠️ warning here, adding a check will stop the app from breaking but this isn't the full SSR support
>
> We need some more work (can be done in an independent pull request) to render styles from cosmos in the server. Otherwise, you would get a flash of unstyled content until the client JS bundle arrives.
>
> Ref: styled-components.com/docs/advanced#server-side-rendering

core/components/_helpers/globals.js
nicogarcia: hi! 🤔 Are you referring to the new flag I added? I also thought of reusing but that would be a breaking change if I'm not wrong (should have left a comment about that, sorry).
My reasoning:
- COSMOS_DISABLE_RESETS=true => minimum style resets
- COSMOS_DISABLE_RESETS=false => recommended style resets
- COSMOS_MANUAL_RESETS=true => none of the resets applied (to avoid knowing how to do it and for example depending on `document`)
- COSMOS_MANUAL_RESETS=false => include the resets driven by the COSMOS_DISABLE_RESETS variable

I think the COSMOS_DISABLE_RESETS behavior should be more aligned with it's name and have the effect that the MANUAL one has, but I didn't want to break the interface we have in the current version.

If my reasoning is right, feel free to take a decision over breaking or not and I can apply it 🙂

nicogarcia:
Regarding full SSR support, I didn't notice what you pointed in the styled components ref (thanks for that 💪), but I played (partially) with it yesterday and, if I'm not wrong, exporting the styles like in this PR and adding the right documentation on how to complement the styled components SSR code would be enough to support SSR.

This is what I'm thinking about (applied in Next.js, [original example](https://github.com/zeit/next.js/blob/master/examples/with-styled-components/pages/_document.js)):
```js
import Document, { Head, Main, NextScript } from 'next/document'
import { ServerStyleSheet } from '@auth0/cosmos/styled' // Would this work? I'd say so

// Import the resets as string
import { resets } from '@auth0/cosmos'

export default class MyDocument extends Document {
  static getInitialProps ({ renderPage }) {
    const sheet = new ServerStyleSheet()
    const page = renderPage(App => props => sheet.collectStyles(<App {...props} />))
    const styleTags = sheet.getStyleElement()
    return { ...page, styleTags }
  }

  render () {
    return (
      <html>
        <Head>
		  // Basic addition of resets (may be done through props)
		  <style>{resets}</style>
          {this.props.styleTags}
        </Head>
        <body>
          <Main />
          <NextScript />
        </body>
      </html>
    )
  }
}
```

wdyt? makes sense?

nicogarcia:
Hey, I've created a full SSR solution (and tested it, it works! 🚀)

The only missing thing is `cosmos-fonts` has the same `insertAtHead` ([see here](https://github.com/auth0/cosmos-fonts/blob/master/index.js)) approach and I had to copy paste the styles to make it work 😬

I've run out of time for this, but if you agree with this implementation, we should use the same flag in `cosmos-fonts` to avoid using `document` and exporting the fonts so people can use it however they want in their server 🙂

Please let me know how it looks for you!

PR #1462, opened 2019-01-27
#1461 Fix checkbox when label is too long

### LINKED ISSUE #1461: Checkbox visually breaks when label is long
**Describe the bug**
When you stress out the Checkbox, the component visually breaks

**Expected behavior**
The checkbox would handle any text size and create proper indentation so the text never goes below the checkbox square.

Let's add proper testing as well. 

**Screenshots**

Rendered code:
screen shot 2019-01-25 at 19 20 16

Visual reference:
screen shot 2019-01-25 at 19 28 19

**Cosmos version:**
0.14

m4tr1k:
Fixes #1461 
I changed the `const Label` from `span` to `p`, I add `display: flex` to the component and I did some changes to the checkbox itself to be always on top and in the same position, as @landitus wished.

Hope this could solve your problem!

### DIFF
--- a/core/components/atoms/checkbox/checkbox.js
+++ b/core/components/atoms/checkbox/checkbox.js
@@ -6,7 +6,7 @@ import Automation from '../../_helpers/automation-attribute'
 import containerStyles from '../../_helpers/container-styles'
 
 const CheckMark = styled.span``
-const Label = styled.span``
+const Label = styled.p``
 
 const justifyContent = {
   horizontal: `margin-right: ${spacing.medium}`,
@@ -33,6 +33,7 @@ const Checkbox = props => (
 Checkbox.Option = styled.label`
   ${containerStyles};
 
+  display: flex;
   position: relative;
   cursor: pointer;
   margin-bottom: 0;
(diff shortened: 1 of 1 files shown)


andresgalante:
Hey @m4tr1k thanks a lot, as always this is great! 💥 

Just a few comments:

- The `p` should be a `div` since it's not a paragraph we don't want to have semantic to that element
- We should reproduce the same in radios. They probably have the same issue
- Lets add a stressed test to the storybooks.

If you don't have time to do it, please let me know and I'll patch your PR tomorrow morning.

Thanks ❤️

m4tr1k:
Hi @andresgalante, I would love to do that. I will try to do that now, but maybe I will not have time to finish all the requirements. If I can't finish it now, I will do it tomorrow!

m4tr1k:
@andresgalante radios don't have the same issue, but the radio is centered vertically, and it's not suppose to be that way.
I will commit each requirement, one by one!

m4tr1k:
I don't have time to add those tests to the storybook, I can do it tomorrow if you don't mind!

m4tr1k:
Let me know what you think about my work!
Thank you 😄

andresgalante:
Hey @m4tr1k, this is fantastic! thanks a lot!

Just one thing, the bug was happening on checkboxes outside a group. For example, a stand-alone checkbox to accept the terms and conditions.

It'd be great if we add a story to show that use-case: a single checkbox outside the group with a lot of text.

Would you like to add it yourself? if not let me know and I'll patch the PR.

The rest is good to go!

m4tr1k:
Hi @andresgalante, yes I can do that!

PR #1468, opened 2019-01-30
[DXDP-167] Forms improvements

andresgalante:
This PR refactors forms HTML and CSS to improve the semantics and make it responsive.

I think that the only API change this introduces is the `checkbox` prop when the form field is related to radios or checkboxes. For example:

```jsx
<Form.Field checkbox label="Agree">
  <Radio name="example1" selected="one" onChange={_ => {}}>
     <Radio.Option value="one">One</Radio.Option>
     <Radio.Option value="two">Two</Radio.Option>
   </Radio>
</Form.Field>
```

Under the hood, it doesn't only align things better but also adds accessibility and correct semantics to the output.

Still to be done before merging this PR:

- Make it play nicely with #1439 (@siddharthkp should we merge that one first and rebase?)
- Add aria tags and ids for the error messages to meet [WCAG standards](https://www.w3.org/WAI/tutorials/forms/notifications/) and [this](https://pf4.patternfly.org/components/Form/examples/#accessibility) might help too.
- Test it on Manhattan

On a separate story, we should write really good docs and examples for forms.


siddharthkp:
> Make it play nicely with #1439 (@siddharthkp should we merge that one first and rebase?)

I vote yes, i'll work on it

siddharthkp:
@francocorreasosa @andresgalante Is it possible to read the children and figure out if the children is a Checkbox? 🤔 We might be able to remove the `checkbox` prop which would be cool

francocorreasosa:
@siddharthkp That is a good idea, we should try 👍 cc @andresgalante

andresgalante:
If we can do it magically, then the better!

francocorreasosa:
Done ✨

andresgalante:
@francocorreasosa you are the best!

@landitus can you please review the chromatic changes so we can merge this one?

francocorreasosa:
We need to review the `restricted` vs `fullWidth` strategy with @andresgalante. Both approaches are not satisfying current behaviour in internal tests.

landitus:
I just saw some small issues:
- The spacing between the Label and field in the (top aligned labels) should be 8px.
- There is an [example of a full page](https://www.chromaticqa.com/snapshot?appId=5aba7f6a07f95800202658ff&id=5c76e3ac4bc82d00248334ee) that includes the form that now changes. It would be a good opportunity to use a Layout to constrain the Form.

francocorreasosa:
@landitus We should be ready to go 👍

landitus:
Just a few comments: 
- Check that the actions are not always aligned to the labels on the left.
- On the labels on the left with full-width, let's see if we can cap the width of the label to the same width we used to have (check the screenshot). this way, the label will not grow, only the form controls

screen shot 2019-03-06 at 13 51 23
.

francocorreasosa:
It seems snapshots are broken, fixing...

francocorreasosa:
Done

andresgalante:
@landitus we just need you to approve the chromatic changes and @francocorreasosa to check this on Manhattan to merge.

andresgalante:
@landitus I've made the change so the labels are fixed at 200px and the forms are restricted to 624px instead of 625.

francocorreasosa:
@landitus can we get a re-review of this? Thanks 🙇

core/components/_helpers/component.story.js
```diff
@@ -46,14 +46,14 @@ const InputBox = styled(Box.withComponent('input'))`
...
-    <Box />
-    <Box bg="blue" />
-    <TallBox bg="blue" />
-    <ShortBox bg="blue" />
-    <InputBox bg="blue" defaultValue="okay" />
-    <PasswordBox bg="blue" defaultValue="okay" />
-    <ExtendedPasswordBox bg="blue" defaultValue="okay" />
-  </Stack>
-))
+// storiesOf('Component', module).add('default', () => (
```
francocorreasosa: @andresgalante can we completely remove this if we're not using it?

landitus: Fine by me. It was bugging me seen it as the initial page in Storybook :)

francocorreasosa: Done 👍

francocorreasosa:
:shipit:

PR #1473, opened 2019-01-31
Fix checkbox event firing twice

### LINKED ISSUE #1471: Checkbox's onChange is triggered twice for a click
**Describe the bug**
When we click on the checkbox the onChange is triggered twice

**Expected behavior**
To be triggered only once per user interaction.

**Screenshots**
N/A

**Cosmos version:**
0.14.1

**Additional context**
Seems onChange is called by the onClick of the input & label of the component

francocorreasosa:
Fixes #1471 

We were passing the onChange event to both the option (label) and the checkbox itself. This caused the event to fire two times when the checkbox was toggled, since the label also fires it. Removing the event from the label solves the issue since now the checkbox is the only component providing this event.


landitus:
- Is there any test (better than the one we deleted) we should have for the checkbox?
- @siddharthkp 	can you review?

francocorreasosa:
@landitus We have the new one here: https://github.com/auth0/cosmos/blob/master/internal/test/integration/atoms/checkbox.test.js

landitus:
Thanks @francocorreasosa 	I've missed that :) 

Then it's review and merge!

siddharthkp:
Just saw the `🚨hi priority` on this. We can merge this PR (I've tested that it works) and continue to improve the code.

core/components/atoms/checkbox/checkbox.js
```diff
@@ -14,7 +15,11 @@ const justifyContent = {
 }
 
 const Checkbox = props => (
-  <Checkbox.Option readOnly={props.readOnly} {...Automation('checkbox')} {...props}>
+  <Checkbox.Option
+    readOnly={props.readOnly}
+    {...Automation('checkbox')}
+    {...exclude('onChange', props)}
```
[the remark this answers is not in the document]
francocorreasosa: Nope, we're still passing it to the checkbox itself:
https://github.com/auth0/cosmos/blob/3ee1da38c0ffe14c94285f30c4ff56806c86f27a/core/components/atoms/checkbox/checkbox.js#L29

core/components/atoms/checkbox/checkbox.js
```diff
@@ -18,7 +18,12 @@ const Checkbox = props => (
   <Checkbox.Option
     readOnly={props.readOnly}
     {...Automation('checkbox')}
-    {...exclude('onChange', props)}
+    {
+      // Prevent passing the onChange handler
+      // both to the label and the checkbox.
+      // See: https://github.com/auth0/cosmos/pull/1473
+      ...exclude('onChange', props)
```
siddharthkp: 👎 Still no clue what's going on 

> See: https://github.com/auth0/cosmos/pull/1473

lol @ recursion

Repeating myself:

> Just saw the `🚨hi priority` on this. We can merge this PR (I've tested that it works) and continue to improve the code.

francocorreasosa: Commented it in the body of the PR. The recursion is intentional and is intended for future reference, as we've been doing since a couple of months.

siddharthkp: Oops sorry, didn't read the description. Honest mistake

francocorreasosa: np, added it a couple minutes ago. It wasn't there before.

siddharthkp:
> We were passing the onChange event to both the option (label) and the checkbox itself.

Should we pass props down to label at all? The label is just text, any additional props given by the user are meant for the input

francocorreasosa:
Okay, sounds good. Let's keep just `readOnly` and `id` then.

siddharthkp:
Does the `id` get applied to both label and input? If yes, we should pick one - the input I think

francocorreasosa:
Done. Merging 👍

PR #1477, opened 2019-02-04
Go deeper on SSR solution

nicogarcia:
### Briefing
This PR bugfixes previous solution and adds an export of ServerStyleSheet of the styles-components library


francocorreasosa:
Since this PR is not pointing directly to `master` I'm going to merge it to `1456-fix-ssr` and we can continue reviewing in that #1457.

PR #1488, opened 2019-02-05
Consistent appearance prop

siddharthkp:
There are 4 components with customisation appearance:

| Component        | prop           | values  |
| ------------- |-------------| -----|
| Alert | type | ['default', 'information', 'success', 'warning', 'danger'] |
| Badge | appearance | ['default', 'information', 'success', 'warning', 'danger'] |
| Label | appearance | ['default', 'information', 'success', 'warning', 'danger'] |
| Button | appearance | ['default', 'primary', 'secondary', 'cta', 'link', 'destructive'] |

The Alert matches the other components in values but not in prop name.

&nbsp;

**This PR fixes that by renaming `type` to `appearance` (with soft deprecation)**

(added story for old API as well to make sure we don't break it prematurely)

&nbsp;

The button doesn't fall into the same category of options, I don't think there is value in forcing it to be consistent either.

So, I'm not sure what the course of action should be here:

1. Accept it as an exception and leave it there. At least, the prop is called `appearance`. The prop type validations help in pointing out that the same values don't apply to Button.

2. Change the prop name (`appearance` to `purpose`?) so that the exception is more visible - that removes expectations of consistency. We can enforce this further by adding a prop type validation that warns users if they even try to use appearance

I'm more inclined towards option 1 but want to hear your opinions

(Will add to migration guide as well if the change is accepted: https://github.com/auth0/cosmos/pull/1490)


francocorreasosa:
Looks good 👍

Regarding buttons case, I'd go with option 1 too since Button in a massively used component and the impact of changing its API now would be unnecessarily high.

siddharthkp:
> Button in a massively used component and the impact of changing its API now would be unnecessarily high.

Well, it's now or v2 😄. As always, we should do a soft deprecation + give codemods.

We should grab this chance to make sweeping API chance. Worst case - we'll do a find and replace and submit a PR 😄

francocorreasosa:
Yeah I know, just thinking is maybe not worth the massive change vs the value it could provide.

siddharthkp:
> Yeah I know, just thinking is maybe not worth the massive change vs. the value it could provide.

I feel you.

This is a bigger question for all the API changes we want to do. 

Future code benefits from a cleaner API. It's a pain to refactor existing code because there is little return on investment (the app will basically work the same, but be easier to maintain - long term return)

We should still make the changes of course. But, make the migration in a batch, preferably with bots. (See https://github.com/auth0/cosmos/pull/1490)

landitus:
Catching up on this, I'll review it.

francocorreasosa:
Just did a backmerge from master to keep this branch up to date and avoid possible merge conflicts in the future.

PR #1490, opened 2019-02-06
Start migration guide

siddharthkp:
Staring a migration guide that we can keep adding to.

Let's merge this and add stuff as we go.


landitus:
Lovely

PR #1498, opened 2019-02-08
Fix sidebar linkgroup item colored when open but no subitems selected

francocorreasosa:
Do not paint current sidebar linkgroup item link when the linkgroup is open but no subitem is selected.


francocorreasosa:
Accepting same fix I already accepted in #1497 which for some reason Chromatic does not recognize as already accepted.

PR #1500, opened 2019-02-08
Fix sidebar not changing styles when selected link prop changed

francocorreasosa:
https://auth0-cosmos-franco-sidebar-props-fix.now.sh/sandbox/?selectedKind=Sidebar&selectedStory=interactive%20example&full=0&addons=1&stories=1&panelRight=0


landitus:
Looks good to me, can you check @jcenturion ?

PR #1501, opened 2019-02-08
Use Popper's positionFixed mode for tooltips

francocorreasosa:


landitus:
@francocorreasosa 	I can't see the example, seems it's not being rendered by chromatic.

francocorreasosa:
This is strange, we may need to restart the pipeline

landitus:
Now it looks great to me !! I forgot we have a special helper to render the dialog in Storybook. thanks @francocorreasosa 	!

PR #1505, opened 2019-02-12
Image API: support src instead

siddharthkp:
```diff
- <Image source="./image_url" />
+ <Image src="./image_url" />
```


siddharthkp:
Chromatic changes shows tooltip with id changes 🤷‍♂️ Looks like some old snapshots were merged onto master

PR #1522, opened 2019-02-21
Tooltip for actions

francocorreasosa:
You can use a special variation of the tooltip which is made for input actions.

It does recieve a `content` object with the following keys:

- `default`: Is the default text to show in the tooltip (same as `content` in a normal tooltip).
- `loading`: Is the text will show if the child button `onClick` handler is an async function (Promise) while it's loading.
- `success`: Is the text will show when the `onClick` returns (if it's a sync function) or resolves (if it's async).
- `error`: If the text will show if the `onClick` is an async function and is rejected.

#### Example

```jsx
<TextInput
  type="text"
  placeholder="Enter some text"
  actions={[
    <Tooltip.Action
      content={{
        default: 'Copy',
        success: 'Copied!'
      }}
    >
      <Button icon="copy" onClick={e => console.log(e)} />
    </Tooltip.Action>,
    <Tooltip.Action
      content={{
        default: 'Do work',
        loading: 'Working...',
        success: 'Done!',
        error: 'Could not work'
      }}
    >
      <Button icon="wrench" onClick={sampleAsyncAction} />
    </Tooltip.Action>,
    <Tooltip.Action
      content={{
        default: 'Do work',
        loading: 'Working...',
        success: 'Done!',
        error: 'Could not work'
      }}
    >
      <Button icon="reload" onClick={sampleAsyncFailingAction} />
    </Tooltip.Action>,
    <Button icon="delete" onClick={e => console.log(e)} />
  ]}
/>
```

demo

### DIFF
--- a/core/components/atoms/tooltip/action.js
+++ b/core/components/atoms/tooltip/action.js
@@ -52,7 +52,7 @@ class ActionTooltip extends React.Component {
    */
   setTooltipContent(newContent, { resetDelay = 3000 } = {}, additionalState = {}) {
     const content = this.preprocessContent()
-    this.setState({ tooltipContent: newContent })
+    this.setState({ tooltipContent: newContent, ...additionalState })
 
     this.clearTimer()
 
--- a/core/components/atoms/tooltip/action.js
+++ b/core/components/atoms/tooltip/action.js
@@ -59,6 +59,8 @@ class ActionTooltip extends React.Component {
       return
     }
 
+    const { resetDelay } = this.props
+
     this.timer = setTimeout(
       () => this.setState({ tooltipContent: content.default }),
       resetDelay
--- a/core/components/atoms/tooltip/action.js
+++ b/core/components/atoms/tooltip/action.js
@@ -79,7 +79,7 @@ class ActionTooltip extends React.Component {
     const newHandler = event => {
       this.resetState()
       if (content.loading) {
-        this.setTooltipContent(content.loading, { resetDelay: resetDelay }, { taskIsRunning: true })
+        this.setTooltipContent(content.loading, { taskIsRunning: true })
       } else {
         this.setState({ taskIsRunning: true })
       }
--- a/core/components/atoms/tooltip/action.js
+++ b/core/components/atoms/tooltip/action.js
@@ -6,6 +6,7 @@ class ActionTooltip extends React.Component {
   constructor(props) {
     super(props)
     this.state = { tooltipContent: null }
+    this.timer = null
   }
 
   /**
@@ -27,7 +28,12 @@ class ActionTooltip extends React.Component {
   setTooltipContent(newContent, { resetDelay = 3000 } = {}) {
     const content = this.preprocessContent()
     this.setState({ tooltipContent: newContent })
-    setTimeout(() => this.setState({ tooltipContent: content.default }), resetDelay)
+
+    if (this.timer) {
+      clearTimeout(this.timer)
+    }
+
+    this.timer = setTimeout(() => this.setState({ tooltipContent: content.default }), resetDelay)
   }
 
   /**
--- a/core/components/atoms/tooltip/action.js
+++ b/core/components/atoms/tooltip/action.js
@@ -49,7 +49,7 @@ class ActionTooltip extends React.Component {
         .then(result => {
           this.setTooltipContent(content.success)
         })
-        .catch(() => this.setTooltipContent(content.errored || 'Something did not work'))
+        .catch(() => this.setTooltipContent(content.error || 'Something did not work'))
     }
 
     return React.cloneElement(button, { onClick: newHandler, appearance: 'link' })
@@ -73,7 +73,7 @@ ActionTooltip.propTypes = {
       default: PropTypes.string.isRequired,
       loading: PropTypes.string,
       success: PropTypes.string,
-      default: PropTypes.string
+      error: PropTypes.string
     })
   ])
 }
--- a/core/components/atoms/tooltip/action.js
+++ b/core/components/atoms/tooltip/action.js
@@ -5,7 +5,7 @@ import Tooltip from './tooltip'
 class ActionTooltip extends React.Component {
   constructor(props) {
     super(props)
-    this.state = { tooltipContent: null }
+    this.state = { tooltipContent: null, asyncTaskIsRunnning: false }
   }
 
   /**
@@ -57,7 +57,7 @@ class ActionTooltip extends React.Component {
 
   render() {
     const { tooltipContent } = this.state
-    const { content, actionSuccess } = this.props
+    const content = this.preprocessContent()
     const children = this.processAction()
 
     return (
--- a/core/components/atoms/tooltip/action.js
+++ b/core/components/atoms/tooltip/action.js
@@ -5,7 +5,7 @@ import Tooltip from './tooltip'
 class ActionTooltip extends React.Component {
   constructor(props) {
     super(props)
-    this.state = { tooltipContent: null }
+    this.state = { tooltipContent: null, taskIsRunning: false }
     this.timer = null
   }
 
@@ -13,9 +13,23 @@ class ActionTooltip extends React.Component {
     this.clearTimer()
   }
 
+  /**
+   * Resets the tooltip content timer if appropiate.
+   */
   clearTimer() {
     if (this.timer) {
       clearTimeout(this.timer)
+    }
+  }
+
+  /**
+   * Resets the tooltip content timer and sets
+   * its content to the default state.
+   */
+  resetState() {
+    if (!this.state.taskIsRunning) {
+      this.clearTimer()
+      this.setState({ tooltipContent: null })
     }
   }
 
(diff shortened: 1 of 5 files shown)


core/components/atoms/tooltip/action.js
```diff
@@ -0,0 +1,81 @@
...
+    return React.cloneElement(button, { onClick: newHandler, appearance: 'link' })
+  }
+
+  render() {
+    const { tooltipContent } = this.state
+    const { content, actionSuccess } = this.props
+    const children = this.processAction()
+
+    return (
+      <Tooltip content={tooltipContent ? tooltipContent : content.default} children={children} />
```
cristiandouce: if `props.content` can be a string, then `content.default` could end up returning undefined, right?

core/components/atoms/tooltip/action.js
```diff
@@ -0,0 +1,81 @@
...
+        })
+        .catch(() => this.setTooltipContent(content.error || 'Something did not work'))
+    }
+
+    return React.cloneElement(button, { onClick: newHandler, appearance: 'link' })
+  }
+
+  render() {
+    const { tooltipContent } = this.state
+    const { content, actionSuccess } = this.props
```
cristiandouce: `actionSuccess` is not defined on `PropTypes`.
also, `content` should be `this.processContent()`, right?
```suggestion
    const content = this.processContent()
```

core/components/atoms/tooltip/tooltip.js
```diff
@@ -94,6 +95,15 @@ class Tooltip extends React.Component {
   }
 }
 
+Tooltip.ObservedElement = class extends React.Component {
```
cristiandouce: 💇‍♂️

core/components/atoms/tooltip/tooltip.md
```diff
@@ -44,3 +44,54 @@ The Tooltip wraps long text by default to make it readable. Try to keep the cont
...
+    </Tooltip.Action>,
+    <Tooltip.Action
+      content={{
+        default: 'Do work',
+        loading: 'Working...',
+        success: 'Done!',
+        error: 'Could not work'
+      }}
+    >
+      <Button icon="reload" onClick={sampleAsyncFailingAction} />
```
cristiandouce: could we have an example where the button has a disabled/loading state as well?

core/components/atoms/tooltip/action.js
francocorreasosa: Good catch ✅

core/components/atoms/tooltip/action.js
```diff
@@ -0,0 +1,81 @@
...
+    const content = this.preprocessContent()
+
+    const newHandler = event => {
+      if (content.loading) {
+        this.setTooltipContent(content.loading, { resetDelay: loadingTimeout || 5000 })
+      }
+
+      Promise.resolve(button.props.onClick(event))
+        .then(result => {
+          this.setTooltipContent(content.success)
```
cristiandouce: what if this is undefined?
I'd probably avoid this, or just use `content.default` as fallback

core/components/atoms/tooltip/action.js
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * Sets the tooltip inner text to `newContent`
+   * and resets the content to `props.content.default` after `resetDelay`.
+   * @param {string} newContent - Text to be set in the tooltip
+   * @param {string} resetDelay - Time in milliseconds to reset back to default text.
+   */
+  setTooltipContent(newContent, { resetDelay = 3000 } = {}) {
+    const content = this.preprocessContent()
+    this.setState({ tooltipContent: newContent })
+    setTimeout(() => this.setState({ tooltipContent: content.default }), resetDelay)
```
cristiandouce: I'd save this timeout id somewhere if possible, and on `componentWillUnmount` just reset everything?

cristiandouce: should we handle something on `onMouseLeave` or whatever trigger we have for the tooltip, cancel this timeout?

core/components/atoms/tooltip/action.js
francocorreasosa: Done ✅

core/components/atoms/tooltip/tooltip.md
francocorreasosa: Done ✅

francocorreasosa:
@landitus hmm 🤔 can't reproduce here. happy to pair with you tomorrow so we can take a look in your computer together.

kapture 2019-02-21 at 20 40 15

siddharthkp:
👋 API suggestion:

`Tooltip.Action` sounds like it renders the button too

Should we call it `Tooltip.ActionWrapper` which conveys that you use it with Actions?

francocorreasosa:
@siddharthkp:

R1. That's fair enough, let's do it 👍 

R2. This solution is not meant to be used with switches, just buttons. I haven't seen any example of an input using a switch inside actually 🤔

core/components/atoms/tooltip/action.js
```diff
@@ -0,0 +1,141 @@
...
+    this.setState({ tooltipContent: newContent })
+
+    this.clearTimer()
+
+    if (newContent === content.default) {
+      return
+    }
+
+    this.timer = setTimeout(
+      () => this.setState({ tooltipContent: content.default, ...additionalState }),
```
cristiandouce:
```suggestion
      () => this.setState({ tooltipContent: content.default }),
```

core/components/atoms/tooltip/action.js
```diff
@@ -0,0 +1,141 @@
...
+  /**
+   * Sets the tooltip inner text to `newContent`
+   * and resets the content to `props.content.default` after `resetDelay`.
+   * @param {string} newContent - Text to be set in the tooltip
+   * @param {string} resetDelay - Time in milliseconds to reset back to default text.
+   * @param {object} additionalState - Any additional data you want to set in the state.
+   */
+  setTooltipContent(newContent, { resetDelay = 3000 } = {}, additionalState = {}) {
+    const content = this.preprocessContent()
+    this.setState({ tooltipContent: newContent })
```
cristiandouce:
```suggestion
    this.setState({ tooltipContent: newContent, ...additionalState })
```

core/components/atoms/tooltip/action.js
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  /**
+   * Sets the tooltip inner text to `newContent`
+   * and resets the content to `props.content.default` after `resetDelay`.
+   * @param {string} newContent - Text to be set in the tooltip
+   * @param {string} resetDelay - Time in milliseconds to reset back to default text.
+   * @param {object} additionalState - Any additional data you want to set in the state.
+   */
+  setTooltipContent(newContent, { resetDelay = 3000 } = {}, additionalState = {}) {
```
cristiandouce:
```suggestion
  setTooltipContent(newContent, additionalState = {}) {
```

core/components/atoms/tooltip/action.js
```diff
@@ -0,0 +1,141 @@
...
+  preprocessContent() {
+    const rawContent = this.props.content
+    return typeof rawContent === 'string' ? { default: rawContent } : rawContent
+  }
+
+  /**
+   * Sets the tooltip inner text to `newContent`
+   * and resets the content to `props.content.default` after `resetDelay`.
+   * @param {string} newContent - Text to be set in the tooltip
+   * @param {string} resetDelay - Time in milliseconds to reset back to default text.
```
cristiandouce:
```suggestion
```

core/components/atoms/tooltip/action.js
```diff
@@ -0,0 +1,141 @@
...
+    const content = this.preprocessContent()
+    this.setState({ tooltipContent: newContent })
+
+    this.clearTimer()
+
+    if (newContent === content.default) {
+      return
+    }
+
+    this.timer = setTimeout(
```
cristiandouce:
```suggestion
    const { resetDelay } = this.props

    this.timer = setTimeout(
```

core/components/atoms/tooltip/action.js
```diff
@@ -0,0 +1,141 @@
...
+   * from a text input / text area action.
+   */
+  processAction() {
+    const { children: button, resetDelay } = this.props
+    const content = this.preprocessContent()
+
+    const newHandler = event => {
+      this.resetState()
+      if (content.loading) {
+        this.setTooltipContent(content.loading, { resetDelay: resetDelay }, { taskIsRunning: true })
```
cristiandouce:
```suggestion
        this.setTooltipContent(content.loading, { taskIsRunning: true })
```

cristiandouce:
> 👋 API suggestion:
> 
> Q1. Should we call it `Tooltip.ActionWrapper` which conveys that you use it with actions? `Tooltip.Action` sounds like it renders the button too. The idea is to communicate you bring your own actions.
> 
> Q2: Does this assume/enforce the child is a `Button` with `onClick` or are we open to accepting other components like `Switch` with `onChange`?

Q1: We should bring this up with the crew. I personally like `Tootip.Actions`.

Q2: We could discuss expanding the support in the future. Now it's just a very specific implementation to unblock us.

PR #1527, opened 2019-02-25
[DXDP-175] Fix select's dialog overflow

francocorreasosa:
### Before
imagen

### After
imagen

### DIFF
--- a/core/components/atoms/select/select.js
+++ b/core/components/atoms/select/select.js
@@ -125,6 +125,13 @@ const oneOrMore = options => {
   return transformation(options)
 }
 
+const componentOverrides = {
+  MultiValue: cosmosMultiValueTagRenderer,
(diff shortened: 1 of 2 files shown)


core/components/atoms/select/select.js
```diff
@@ -124,56 +125,94 @@ const oneOrMore = options => {
...
+
+  render() {
+    const props = this.props
+    /*
+      select boxes do not support readonly like input boxes,
+      but they do have disabled. we need the style of readOnly input
+      and functionality of select disabled
+    */
+
+    const componentOverrides = {
```
cristiandouce: if this is static we could move it out of the `render()` scope, so it doesn't re-generate each time render is called.

francocorreasosa: Done ✨

PR #1528, opened 2019-02-25
[DXDP-183] Conditional select

francocorreasosa:
Only use the new select if using `multiple` or `searchable`

- [x] Check the inline select is working again (MH)
- [x] Native: Render the placeholder conditionally.
- [x] Missing the hasError prop
- [x] Storybook: Check all examples appear in Storybook.


landitus:
QQ: I think I remember the multi-select showed the selected items in the drawner as well as in the input. Now it shows the remaining items only. Is this something we can control?

francocorreasosa:
This has been always be this way AFAIK. We can take a look at react-select's docs to see if this is configurable.

PR #1536, opened 2019-03-03
Update stack-layout.js

7IHd:
Fix spelling


francocorreasosa:
Thanks for the fix @7IHd! Will move this to a new internal branch so the CI can work fine and we can merge 👍

PR #1557, opened 2019-03-13
[DXDP-240] Avoid passing `title` to component's root elements

francocorreasosa:


core/components/_helpers/root-props.js
```diff
@@ -0,0 +1,17 @@
...
+ *
+ *      <Dialog title="This is a title"> ... </Dialog>
+ *
+ * If we just do {...props} in the root element of the component, you
+ * get the `title` prop passed to the native HTML element, and this
+ * causes a tooltip to appear in the element when you rest the cursor on it.
+ *
+ * @param {object} props
+ */
+export const rootProps = props => {
```
hzalaz: can we add a test at least for this?

francocorreasosa: Done 👍

PR #1560, opened 2019-03-14
 Fix submenu not being open when subitem is selected 

francocorreasosa:
plus use `shallow` instead of `render` in Navigation snapshot tests.

### DIFF
--- a/core/components/molecules/navigation/navigation-item.js
+++ b/core/components/molecules/navigation/navigation-item.js
@@ -10,6 +10,7 @@ import NavigationSubnav from './navigation-subnav'
 import { childrenMover } from '../../_helpers/children-mover'
 
 const ariaCurrent = props => (props.selected ? { 'aria-current': 'page' } : {})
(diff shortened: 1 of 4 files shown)


core/components/molecules/navigation/navigation-item.js
```diff
@@ -48,7 +48,7 @@ const processIcon = (children, selected, inSubMenu) =>
 class NavigationItem extends React.Component {
   constructor(props) {
     super(props)
-    this.state = { subMenuOpen: props.defaultOpen || false }
+    this.state = { subMenuOpen: props.defaultOpen || null }
```
cristiandouce: why don't we default the initial state value from the `selected` props?
we could create a method getter...

```js
get isSelected() {
  const { selected, subMenu } = this.props;
  return selected || findSelectedSubItem(subMenu);
}

// and then just call
this.state = { subMenuOpen: defaultOpen || this.isSelected };
// ...
```

PR #1561, opened 2019-03-15
Fix avatarUrl helper, add tests

francocorreasosa:


core/components/_helpers/avatar-url.js
```diff
@@ -9,6 +9,9 @@ const urlTemplate = (hash, initials, size) =>
   `https://s.gravatar.com/avatar/${hash}?s=${size}&r=pg&d=${initialsAvatar(initials)}`
 
 export default function avatarUrl(email, initials, size = 480) {
-  const hash = md5(email.toLowerCase())
+  if (!email && !initials)
+    throw new Error('You need to provide either `email`, `initials` or both to the Avatar.')
```
cristiandouce: isn't that why we do have a default profile svg?

internal/test/unit/helpers/avatar-url.test.js
```diff
@@ -0,0 +1,38 @@
...
+      'https://s.gravatar.com/avatar/23463b99b62a72f26ed677cc556c44e8?s=480&r=pg&d=https%3A%2F%2Fcdn.auth0.com%2Favatars%2Ffc.png'
+    )
+  })
+
+  it('to return a custom size URL if size is provided', () => {
+    const url = avatarUrl('example@example.com', 'FC', 368)
+    expect(url).toEqual(
+      'https://s.gravatar.com/avatar/23463b99b62a72f26ed677cc556c44e8?s=368&r=pg&d=https%3A%2F%2Fcdn.auth0.com%2Favatars%2Ffc.png'
+    )
+  })
```
cristiandouce: ❤️

core/components/_helpers/avatar-url.js
francocorreasosa: That case is handled by the Avatar component itself (by rendering a local SVG, so not a URL).

francocorreasosa: Nope, just checked. This is a CDN URL, let me change this so we can centralize it :)

francocorreasosa: Done

PR #1569, opened 2019-03-20
Improves List docs

andresgalante:
This PR improves the docs for the list with new examples.


andresgalante:
@francocorreasosa is it worth marking the label as deprecated now or on a new PR?

Issue #1542, opened 2019-03-07
Documentation: Select component's doc has duplicate options props


ubenzer:
**Describe the bug**
Please see the screenshot. Two `options` are shown in the doc.

**Expected behavior**
Only one `options` is shown in the docs.

**Screenshots**
![](https://user-images.githubusercontent.com/396400/53965257-22ffdb80-40f1-11e9-86b4-655c197001e9.png)

**Cosmos version:**
0.19.0

**Additional context**
N/A

landitus:
Done !

francocorreasosa:
We have not fixed this

landitus:
ups!

cliffpyles:
Hi @francocorreasosa @landitus ! This is my first contribution to the project, and this one seemed like it was pretty straightforward. Let me know if I need to make any other changes though.

PR #1541, opened 2019-03-07
Fixes the duplicate options in the docs. (#1540)

francocorreasosa:
via @ubenzer 

screen shot 2019-03-07 at 3 52 26 pm

PR #1577, opened 2019-03-28
Fix typo in Select documentation

cliffpyles:
This fixes a typo that was causing an extra prop to be shown in the code snippet for the Select. #1542


francocorreasosa:
Hey @cliffpyles 👋Thank you for taking the time to send a PR. Unfortunately, this does not solve the issue because removing `{props}` from markdown causes the prop table to be unable to pass props to the component.

Related to https://github.com/auth0/cosmos/pull/1541

cliffpyles:
@francocorreasosa Got it! I'll close this and take another look then. Thanks for the feedback!

francocorreasosa:
Sure, thank you.

PR #1579, opened 2019-04-08
Changed tv icon

landitus:
The TV icon got confused with a computer monitor, so now it's tweaked to look more like a Smart TV.

Before/After
Screen Shot 2019-04-08 at 11 38 28


francocorreasosa:
Let's merge after #1576

landitus:
Now it's fixed

PR #1602, opened 2019-05-29
Removes tokens definitions on the guidelines

andresgalante:
Until we implement the token system, let's remove the guidelines


internal/docs/guides/contribution-guide/system-guidelines.js
```diff
@@ -125,6 +125,10 @@ const SystemGuidelines = () => (
     <Heading3>Theming</Heading3>
 
     <Text>
+      We are working on a 2 layer theming system to open Cosmos for global and glanular
```
loginist:
```suggestion
      We are working on a 2 layer theming system to open Cosmos for global and granular
```

PR #1636, opened 2019-07-04
Fix syntax error in the Alert component doc

pdillon:
By submitting a PR to this repository, you agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md). Please see the [contributing guidelines](https://github.com/auth0/.github/blob/master/CONTRIBUTING.md) for how to create and submit a high-quality PR for this repo.

### Description

This PR addresses the syntax error currently rendered at https://auth0-cosmos.now.sh/docs/#/component/alert

image

The parsing in prop-string.js is handled by the default case, which wraps all values in `{}`-  breaking on string types. Added a ReactNode case to check the data type, similar to other cases. Considered extending the default case to handle multiple types, but opted for minimal changes to remove the error first.

### Testing
Unit tests pass locally. 
No additional or updated tests needed - component functionality was not altered.

### Checklist
- Verified local build correctly renders the Alert example.
image
- All active GitHub checks for tests, formatting, and security are passing.
- The correct base branch is being used, if not `master`.

PR #1662, opened 2019-08-29
Fix syntax error in the Alert component doc (#1636)

pdillon:
_( Copied from #1636 )_
----------------------------
### Description

This PR addresses the syntax error currently rendered at https://auth0-cosmos.now.sh/docs/#/component/alert

image

The parsing in prop-string.js is handled by the default case, which wraps all values in `{}`-  breaking on string types. Added a ReactNode case to check the data type, similar to other cases. Considered extending the default case to handle multiple types, but opted for minimal changes to remove the error first.

### Testing
Unit tests pass locally. 
No additional or updated tests needed - component functionality was not altered.

### Checklist
- Verified local build correctly renders the Alert example.
image
- All active GitHub checks for tests, formatting, and security are passing.
- The correct base branch is being used, if not `master`.

### DIFF
--- a/core/components/atoms/alert/alert.tsx
+++ b/core/components/atoms/alert/alert.tsx
@@ -18,7 +18,7 @@ export interface IAlertProps {
   type?: IAlertAppearance // deprecated: use appearance
   appearance?: IAlertAppearance
   icon?: string
-  title?: string | React.ReactNode
+  title?: React.ReactNode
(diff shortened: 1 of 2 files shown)


loginist:
PR preview at: https://auth0-cosmos-pr-1636.now.sh/docs/#/component/alert

PR #1674, opened 2019-10-12
fix in button props comment

solankivj94:
#### Fix button `iconAlign` prop comment

Before: Name of icon

After: Align Icon Left or Right


loginist:
Thanks!

PR #1678, opened 2019-10-23
Extra rendering in formgroup fix

not-arzuaga:
### Description

Simple fix for the extra rendering issue when sending and element such as null or undefined on a form group.

### Testing

To test this fix, when creating a formGroup in the sandbox you can send null or an undefined element on it to check if they render.

### Checklist

- [ ] I have added documentation for new/changed functionality in this PR or in auth0.com/docs
- [X] All active GitHub checks for tests, formatting, and security are passing
- [X] The correct base branch is being used, if not `master`


loginist:
Thank you @not-arzuaga! 😄

PR #1688, opened 2020-01-06
Add missing dependency in published package

loginist:
Add missing dependency `lodash.isempty` in published package.


loginist:
Messed up with the branch name, closing in favor of #1689

PR #1698, opened 2020-02-10
Makes alerts accessible

andresgalante:
This PR is making Alerts accessible by:

- adding a `role="alert"` to the wrapper
- adding `aria-label="Close"` to the close action
- making the close btn a `button` instead of an anchor


andresgalante:
I am closing this PR since it's already done in #1651

PR #1733, opened 2021-01-13
DXDP-1904 - Fixing Avatar component infinite loop when image not provided and gravatar fails to load

cristiandouce:
#### Description
We've identified an infinite loop in the Avatar component for when the `image` prop is not provided. In combination with bad `initials` or `email` props that end up failing to load, the following line get's HIT:

https://github.com/auth0/cosmos/blob/f96aaee5d6586a71d05c7a80b3cc2c003a4802eb/core/components/atoms/avatar/avatar.tsx#L143

You can notice that there is no handled case for when the Gravatar source has failed to load (except for the one in combination with image errors). This combination of events start an infinite loop of Line 143 being constantly hit, running several requests per second attempting indefinitely to load Gravatar again and again.

#### Motivation and Context
This is a bug which can cause systems to exhaust or block the user due to the excess of requests produced by the infinite loop.

There are stories in the storybook that reproduced the issue, now fixed!

### DIFF
--- a/core/components/atoms/avatar/avatar.tsx
+++ b/core/components/atoms/avatar/avatar.tsx
@@ -33,7 +33,12 @@ const imageForAvatar = (source, handleError) => {
       fit="cover"
       src={source}
       onError={(event) => {
-        event.target.src = null
+        try {
+          event.target.removeAttribute('src')
+        } catch (e) {
+          event.target.src = undefined
+        }
+        event.target.onerror = undefined
         event.target.onError = undefined
         handleError(event)
       }}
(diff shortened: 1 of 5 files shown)


core/components/atoms/avatar/avatar.tsx
```diff
@@ -137,14 +142,15 @@ class Avatar extends React.Component<IAvatarProps, IAvatarState> {
     const { imageErrored, gravatarErrored } = this.state
     const { email, initials, icon, image } = this.props
 
+    // NOTE: The following order of checks MATTER. They determine the fallback
+    //       precedence when there are any missing props or failed attempts.
+    //       Carefully review all scenarios are covered when changing the
+    //       following lines.
```
degrammer, line 148: maybe adding a unit test for this function can prevent this from breaking the logic in the future

degrammer (dismissed):
LGTM

degrammer (approved):
LGTM, great to see the test !!

PR #1756, opened 2022-02-22
Update to use Node 16

dmiller9911:


package.json
```diff
@@ -87,6 +87,6 @@
     "typescript-tslint-plugin": "0.5.3"
   },
   "engines": {
-    "node": "12"
+    "node": "^14||^16"
```
mgwedd, line 90: Do we need to specify an npm version here so that the lockfile version is compatible? Not sure with yarn tbh...

mgwedd (approved):
LGTM — left one question on the project `engines` spec

package.json
dmiller9911: I think this is ok? I couldn't find much on specifying one without the other. Most are around just the node version. If we encounter problems we can add it in later on.

